PR #1, opened 2015-02-27
Facility and provider entities

pmuchowski:
Change-Id: I9af1ef1b4ea9d83eb2ba4b98b4e9bed96adac086


pgesek:
Use the gerrit server for code review instead of github pull requests. Since gerrit replication is a one way street, changes made directly through github will get lost.

kkaczmarczyk:
@pgesek Thanks for info Pawel, thought that dev branches work different way. We will use Gerrit!

PR #3, opened 2015-09-14
MOTECH-1732: Improves datanucleus performance turning off detachAllOn…

pgesek:
…Commit property

Change-Id: I308040414996d9db61b4f12f447c50a27c09614d

### DIFF
--- a/appointments/src/main/java/org/motechproject/appointments/service/impl/AppointmentServiceImpl.java
+++ b/appointments/src/main/java/org/motechproject/appointments/service/impl/AppointmentServiceImpl.java
@@ -10,6 +10,7 @@ import org.motechproject.appointments.reminder.AppointmentReminderService;
 import org.motechproject.appointments.service.AppointmentService;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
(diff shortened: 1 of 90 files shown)


appointments/src/main/java/org/motechproject/appointments/service/impl/AppointmentServiceImpl.java
```diff
@@ -37,21 +38,27 @@
      * @return Copy of the appointment object from the data store
      */
     @Override
+    @Transactional
     public Appointment addAppointment(Appointment appointment) {
```
pgesek, line 42: This should be left as it was

appointments/src/test/java/org/motechproject/appointments/service/it/AppointmentServiceBundleIT.java
```diff
@@ -57,7 +59,7 @@ public void addAppointment() throws Exception {
         assertEquals(result.getExternalId(), current.getExternalId());
     }
 
-    @Test
+    @Test //TODO We cannot add the same two appointmens, because apptId is Unique. Now this test is just updating previous object. Something should change here
```
pgesek: Just expect an exception here for duplicate apptId

PR #6, opened 2015-09-15
MOTECH-1875: Added custom exception to DHIS2 EventHandler

lukskar:
The DataElementNotFoundException was added to dhis2 module to improve error handling
and make error cause clearer to the user.

### DIFF
--- a/dhis2/src/main/java/org/motechproject/dhis2/event/EventHandler.java
+++ b/dhis2/src/main/java/org/motechproject/dhis2/event/EventHandler.java
@@ -44,6 +44,14 @@ public class EventHandler {
     @Autowired
     private DataElementService dataElementService;
 
+    public EventHandler (DhisWebService webService,
+                         TrackedEntityInstanceMappingService trackedEntityInstanceMappingService,
+                         DataElementService dataElementService) {
+        this.dhisWebService = webService;
+        this.trackedEntityInstanceMappingService = trackedEntityInstanceMappingService;
+        this.dataElementService = dataElementService;
+    }
+
     public EventHandler (DhisWebService webService, TrackedEntityInstanceMappingService trackedEntityInstanceMappingService) {
         this.dhisWebService = webService;
         this.trackedEntityInstanceMappingService = trackedEntityInstanceMappingService;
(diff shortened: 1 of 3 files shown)


dhis2/src/main/java/org/motechproject/dhis2/event/EventHandler.java
```diff
@@ -218,6 +219,13 @@ public void handleDataValue(MotechEvent event) {
         Map<String, Object> params = event.getParameters();
 
         DataElement dataElement = dataElementService.findByName((String) params.get(EventParams.DATA_ELEMENT));
+
+        if (dataElement == null) {
+            throw new DataElementNotFoundException("The data element that was sent did not " +
+                    "match any values imported from DHIS2. Please make sure that the data element field matches " +
```
pgesek: Include params.get(EvenParams.DATA_ELEMENT) in the exception message

dhis2/src/main/java/org/motechproject/dhis2/event/EventHandler.java
```diff
@@ -218,6 +219,13 @@ public void handleDataValue(MotechEvent event) {
         Map<String, Object> params = event.getParameters();
 
         DataElement dataElement = dataElementService.findByName((String) params.get(EventParams.DATA_ELEMENT));
+
+        if (dataElement == null) {
```
pgesek, line 218: this could use a unit test

pgesek:
Needs a test & actual details in the messages

dhis2/src/main/java/org/motechproject/dhis2/event/EventHandler.java
```diff
@@ -43,6 +44,14 @@
     @Autowired
     private DataElementService dataElementService;
 
+    public EventHandler (DhisWebService webService,
```
pgesek: We do not need two constructors, this bean should only get constructed with all dependencies. Since autowiring is done at the field level, this class doesn't need constructors at all. So for clarity sake just delete them both.

dhis2/src/test/java/org/motechproject/dhis2/event/EventHandlerTest.java
```diff
@@ -51,6 +57,8 @@
     @Mock
     private TrackedEntityInstanceMappingService trackedEntityInstanceMappingService;
     @Mock
+    private DataElementService dataElementService;
+    @Mock
     private DhisWebService dhisWebservice;
     private DhisStatusResponse response;
     private EventHandler handler;
```
pgesek: Just construct the instance here with the @InjectMocks annotation, and the need for constructors will go away.

PR #27, opened 2015-09-21
MOTECH-1901: Changed way of embedding MDS grids

ngraczewski:
Needs https://github.com/motech/motech/pull/47 to work properly.

### DIFF
--- a/sms/src/main/resources/webapp/index.html
+++ b/sms/src/main/resources/webapp/index.html
@@ -17,19 +17,7 @@
             </li>
         </ul>
     </div>
-    <div class="ui-layout-content">
-        <div class="tab-content" id="tab-content">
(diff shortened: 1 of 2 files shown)


sms/src/main/resources/webapp/index.html
```diff
@@ -17,19 +17,7 @@
...
         </ul>
     </div>
-    <div class="ui-layout-content">
-        <div class="tab-content" id="tab-content">
-            <div id="main-content" class="active">
-                <div ng-view></div>
-            </div>
-        </div>
-    </div>
+    <div embedded-mds-grid></div>
```
pgesek, line 20: any reason for this to be in the index page, not the log page?

pgesek:
What about message campaign?

PR #29, opened 2015-09-22
MOTECH-1411: JDK for Modules set to 1.8

lukskar:
- JDK for Modules is now 1.8 and requires Java8 to compile.
- Changed version of maven-pmd-plugin to 3.5 as old one didn't support JDK 1.8.


pgesek:
test this please

PR #31, opened 2015-09-24
MOTECH-1901: Improved embedded grids in modules

ngraczewski:
Dependent on https://github.com/motech/motech/pull/63


pgesek:
2 comments regarding the functionality:
- I cannot filter on the auto generated date fields in IVR for whatever reason. It works in SMS though.
- The grids in sms and IVR take too little space vertically, if I expand the field i Can that it gets cut off after the first field, and I get a vertical scroll bar for only a part of the window.

ngraczewski:
Newest commit on https://github.com/motech/motech/pull/63 should fix this.

PR #42, opened 2015-10-01
MOTECH-1929: Added the ability to pass extra data with stock tasks

pgesek:
- The query stock ledger actions now contains an extra data field,
  which is a map. Everything from this map will be injected into
  events being fired for each retrieved stock.
- This data can be accessed using the notation {{trigger.key}}, where
  key is the key from the extra data map.


pgesek:
test this please

pgesek:
test this please

sebbrudzinski:
@pgesek OK to merge

PR #44, opened 2015-10-02
Fix for IVR & SMS ITs

pgesek:
Made ITs use our polling client instead of the
simple client. This hopefully will fix the failures
that happen randomly in these tests.


sebbrudzinski:
@pgesek OK to merge

PR #46, opened 2015-10-05
MOTECH-1388: Refactor for Quiz, Course, Chapter and Lesson.

tstalka:
This commit adds description field and properties Map. It Also adds
methods to the MTrainingService for searching by properties Map.


mtraining/src/main/java/org/motechproject/mtraining/domain/CourseUnitMetadata.java
```diff
@@ -36,44 +36,56 @@
     private String content;
 
     /**
-     * Constructor with all arguments
+     * Description of the unit.
+     */
+    @Field
+    @Persistent(defaultFetchGroup = "true")
```
pgesek: Isn't this the default for String fields?

mtraining/src/main/java/org/motechproject/mtraining/service/impl/MTrainingServiceImpl.java
```diff
@@ -174,8 +172,21 @@ public Lesson createLesson(Lesson lesson) {
...
-        return lessonDataService.findLessonByName(lessonName);
+    public List<Lesson> getLessonsByName(String lessonName) {
+        return lessonDataService.findLessonsByName(lessonName);
+    }
+
+    @Override
+    public List<Lesson> getLessonsByProperties(Map<String, String> properties) {
+        List lessons = lessonDataService.executeQuery(new QueryExecution<List<Lesson>>() {
+            @Override
+            public List<Lesson> execute(Query query, InstanceSecurityRestriction restriction) {
```
pgesek: We should create a separate, generic class for this execution, since the body of the method is always the same.

mtraining/src/main/java/org/motechproject/mtraining/service/impl/MTrainingServiceImpl.java
```diff
@@ -133,8 +118,21 @@ public Chapter createChapter(Chapter chapter) {
...
+    public List<Chapter> getChaptersByProperties(Map<String, String> properties) {
+        List chapters = chapterDataService.executeQuery(new QueryExecution<List<Chapter>>() {
+            @Override
+            public List<Chapter> execute(Query query, InstanceSecurityRestriction restriction) {
+                query.setFilter(buildParametersFilter(properties));
+                return (List) QueryExecutor.execute(query, restriction);
+            }
+        });
+
+        return new ArrayList<>(chapters);
```
pgesek: Any reason why we are packing the results of queries into ArrayLists?

pgesek:
@tstalka Needs merge with latest master

PR #47, opened 2015-10-06
MOTECH-1936: Added rounding to the display of decimal fields in IVR

pgesek:
- We round to two decimal points
- We don't display the decimal part if it is equal 0
- If the value is not a number, we leave it as is


sebbrudzinski:
@pgesek OK to merge.

PR #48, opened 2015-10-06
Updated motech.version dependency from 0.26.2 to 0.26.3-SNAPSHOT

craigappl:
The Jenkins release process updated all of the module versions from 0.26.2 to 0.26.3-SNAPSHOT, but missed the motech.version property. This change updates the <motech.version> property from 0.26.2 to 0.26.3-SNAPSHOT.

@pgesek Can you take a quick look at the Jenkins release process and include this?

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -14,7 +14,7 @@
 
     <properties>
         <motech.groupId>org.motechproject</motech.groupId>
-        <motech.version>0.26.2</motech.version>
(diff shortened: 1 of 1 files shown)


pgesek:
@craigappl Yes, I noticed this during the last release but forgot to get on it. From what I can see in the job, we have a step that updates the version, but we seem to be lacking a step that pushes that to GitHub. I will look at updating the job. Thanks.

PR #59, opened 2015-10-14
MOTECH-1514 Added character counter to CMS Lite

martokarski:


cms-lite/src/main/resources/webapp/js/controllers.js
```diff
@@ -40,6 +40,7 @@
             $scope.resourceType = 'string';
             $scope.mode = 'read';
             $scope.select = {};
+            $scope.select.value = '';
```
sebbrudzinski: Can this be done together with the object creation, a line above?

PR #66, opened 2015-10-20
MOTECH-1939 Fixed DHIS2 request payload and response mapping

sebbrudzinski:
The attributes sent to the DHIS2 server were build from the paramters, from the MotechEvent payload, however the DHIS2 server will not accept any attributes, that aren't recognized (created earlier). This caused conflicts when using event handlers in DHIS2. The MotechEvent-specific parameters are now removed before attributes are created.

### DIFF
--- a/dhis2/src/main/java/org/motechproject/dhis2/event/EventHandler.java
+++ b/dhis2/src/main/java/org/motechproject/dhis2/event/EventHandler.java
@@ -61,7 +61,7 @@ public class EventHandler {
     @MotechListener(subjects = EventSubjects.CREATE_ENTITY)
     public void handleCreate (MotechEvent event) {
         LOGGER.debug("Handling CREATE_ENTITY MotechEvent");
-        Map<String, Object> params = new HashMap<>(event.getParameters());
+        Map<String, Object> params = prepareDhisAttributesMap(event.getParameters());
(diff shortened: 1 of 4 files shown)


dhis2/src/main/java/org/motechproject/dhis2/event/EventHandler.java
```diff
@@ -206,84 +286,14 @@ private DhisEventDto createDhisEventFromParams (Map<String, Object> params) {
...
 
-        DataValueSetDto dataValueSetDto = new DataValueSetDto();
-        List<DataValueDto> dataValueDtos = new ArrayList<>();
-        dataValueDtos.add(dataValueDto);
-        dataValueSetDto.setDataValues(dataValueDtos);
-
-        dhisWebService.sendDataValueSet(dataValueSetDto);
+        dhisAttributes.remove(MotechEvent.PARAM_INVALID_MOTECH_EVENT);
+        dhisAttributes.remove(MotechEvent.PARAM_REDELIVERY_COUNT);
+        dhisAttributes.remove(EventParams.MESSAGE_DESTINATION);
```
pgesek, line 295: Can we also remove JOB_ID_KEY from the scheduler service for good measure?

PR #72, opened 2015-10-26
MOTECH-1940: Default UI tab for modules is determined basing on permissions.

lukskar:
For IVR, SMS and mTraining modules:
- Removed AvailabilityController. Now AvailabilityController from MOTECH server-bundle is used.
- Added tabAccessMap. This is a map defining roles as keys and lists of available tabs as values. The tabAccessMap is used instead of defaultURL so the default tab can be determined based on current user permissions.


sms/src/main/resources/META-INF/spring/blueprint.xml
```diff
@@ -50,13 +50,44 @@
...
+                    <key>
+                        <value>send</value>
+                    </key>
+                    <list>
+                        <value>manageSMS</value>
+                    </list>
+                </entry>
+                <entry>
+                    <key>
+                        <value>setting</value>
```
pgesek: I think this should be 'settings' - the tab is broken.

PR #73, opened 2015-10-27
MOTECH-1515 Added support for file:// protocol in IVR outbound service

martokarski:


ivr/src/main/java/org/motechproject/ivr/service/impl/OutboundCallServiceImpl.java
```diff
@@ -119,10 +128,22 @@ public void initiateCall(String configName, Map<String, String> parameters) {
...
+                    generateCallFile(config, params, completeParams, motechCallId);
+                    break;
+                }
+
+                case HTTP_PROTOCOL: {
+                    generateCallHttp(config, params, completeParams, motechCallId);
+                    break;
+                }
+            }
+        } catch (CallInitiationException e) {
```
pgesek: why do we have this catch?

ivr/src/main/java/org/motechproject/ivr/service/impl/OutboundCallServiceImpl.java
```diff
@@ -130,6 +151,32 @@ public void initiateCall(String configName, Map<String, String> parameters) {
...
+        Path path = Paths.get(new URI(myUri));
+        Path tempDirectory = Files.createTempDirectory(path.getParent(), "motech");
+        Path tempPath = Files.createTempFile(tempDirectory, "motech", ".tmp");
+
+        BufferedWriter bufferedWriter = Files.newBufferedWriter(tempPath);
+        for (Map.Entry<String, String> entry : completeParams.entrySet()) {
+            bufferedWriter.write(String.format("%s: %s", entry.getKey(), entry.getValue()));
+            bufferedWriter.newLine();
+        }
+        bufferedWriter.close();
```
pgesek: use the writer in a try-with-resource, closing resources should be done in a finally block

ivr/src/main/java/org/motechproject/ivr/service/impl/OutboundCallServiceImpl.java
```diff
@@ -130,6 +151,32 @@ public void initiateCall(String configName, Map<String, String> parameters) {
...
+
+        Files.move(tempPath, path, StandardCopyOption.ATOMIC_MOVE);
+        tempDirectory.toFile().delete();
+
+        addCallDetailRecord(CallDetailRecord.CALL_INITIATED, config, params, motechCallId);
+    }
+
+    private void generateCallHttp(Config config, Map<String, String> params, Map<String, String > completeParams, String motechCallId) throws IOException {
```
pgesek: rename to executeHttpCall

ivr/src/main/java/org/motechproject/ivr/service/impl/OutboundCallServiceImpl.java
```diff
@@ -130,6 +151,32 @@ public void initiateCall(String configName, Map<String, String> parameters) {
             addCallDetailRecord(CallDetailRecord.CALL_FAILED, config, params, motechCallId);
             throw new CallInitiationException(message, e);
         }
+    }
+
+    private void generateCallFile(Config config, Map<String, String> params, Map<String, String> completeParams, String motechCallId) throws URISyntaxException, IOException {
+        String myUri = mergeUriAndRemoveParams(config.getOutgoingCallUriTemplate(), completeParams);
+
+        Path path = Paths.get(new URI(myUri));
+        Path tempDirectory = Files.createTempDirectory(path.getParent(), "motech");
```
pgesek: use motech-ivr

ivr/src/main/java/org/motechproject/ivr/service/impl/OutboundCallServiceImpl.java
```diff
@@ -130,6 +151,32 @@ public void initiateCall(String configName, Map<String, String> parameters) {
...
+        Path tempPath = Files.createTempFile(tempDirectory, "motech", ".tmp");
```
pgesek: name this callfile.txt

ivr/src/main/java/org/motechproject/ivr/service/impl/OutboundCallServiceImpl.java
pgesek: Do we need a temp directory per file? Just create temp file straight away

ivr/src/test/java/org/motechproject/ivr/it/OutboundCallServiceBundleIT.java
```diff
@@ -139,4 +143,68 @@ public void shouldExecuteJsonRequest() throws IOException {
...
+        configService.updateConfigs(Arrays.asList(config));
+
+        Map<String, String> params = new HashMap<>();
+        params.put("api_key", "qwerty123");
+        params.put("message_id", "123123");
+        params.put("channel", "ivr");
+        params.put("status_callback_url", "http://someUrl.com");
+        params.put("subscribers", "[{\"phone\":\"48700123123\",\"language\":null}]");
+
+        List<String> result = new ArrayList<>();
```
pgesek, line 167: This is pretty much the same as the previous test. Why not just have this one test?

ivr/src/main/java/org/motechproject/ivr/service/impl/OutboundCallServiceImpl.java
martokarski: according to Asterisk wiki (https://wiki.asterisk.org/wiki/display/AST/Asterisk+Call+Files) it is recommended to create a file in another directory on the same filesystem

pgesek: Wait, so we are creating a temp directory inside the asterisk call file directory? We should use the temp directory of the system (/tmp in linux), not the asterisk call directory.

martokarski: Ok, but what if /tmp directory and asterisk call directory are on different filesystems? I don't know if this is a real problem, but this is possible. Problem with different filesystem is that could cause an exception during "atomic" move. I could not use atomic move and create file in temp directory or assume that existence of different filesystems on server is almost impossible.

pgesek: In that case just create a file lock on the output then. We can assume the target system supports it

PR #97, opened 2015-12-11
1926

atish160384:
Removed duplicate datepicker code from modules and added common code in server-bundle.


pgesek:
Closing as a duplicate

atish160384:
In this branch I'm clearing the duplicate code in the modules repository and adding common code in server-bundle. So, I'm creating two pull requests from two different repositories(modules & motech) and they need to be merged simultaneously for it to work

PR #103, opened 2015-12-22
Motech 1926

atish160384:
Removed duplicate datepicker code from modules: commcare & sms

### DIFF
--- a/batch/src/main/resources/application.properties
+++ b/batch/src/main/resources/application.properties
@@ -8,6 +8,6 @@
 #*****************************************************************************
 
 jdbc.driver=${sql.driver}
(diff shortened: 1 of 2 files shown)


sms/src/main/resources/webapp/index.html
```diff
@@ -22,4 +22,72 @@
     <div tab-layout-with-mds-grid></div>
 </div>
 
-<div embedded-mds-filters></div>
+<div id="inner-east" class="sms inner-east ui-layout-pane ui-layout-pane-east">
```
pgesek: What happened here? Why are we no longer embedding filters?

PR #111, opened 2016-01-15
MOTECH-2037: Make motech directory configurable

matkwiatkowski:


testdata/config-locations.properties
```diff
@@ -1 +1 @@
-config.location=${sys:user.home}/.motech/configtest/, ${sys:user.home}/.motech/config/, /etc/motech/
\ No newline at end of file
+config.location=${sys:user.home}/.motech/configtest, ${sys:user.home}/.motech
```
pgesek: This shouldn't change like this, explained in the main repo PR

PR #125, opened 2016-02-05
MOTECH-2139 Added create/update Commcare case Task actions

sebbrudzinski:


commcare/src/main/java/org/motechproject/commcare/domain/CaseJson.java
```diff
@@ -42,101 +42,105 @@
     @SerializedName("case_id")
     private String caseId;
 
-    public String getDateClosed() {
-        return this.dateClosed;
-    }
+    @SerializedName("closed")
```
pgesek, line 45: hard to browse through this, is this just reordering ?

sebbrudzinski, line 45: Yes + a fix of getter name for caseProperties - getProperties() -> getCaseProperties()

PR #127, opened 2016-02-08
MOTECH-1762 Customize the embedded MDS view in IVR

akarczewski:


kmadej:
You shouldn't hide all buttons. for sure lookupButton, fieldsButton, exportButton should be available

PR #130, opened 2016-02-10
MOTECH-1401: SMS - There should be no Call Status Enum

matkwiatkowski:


pgesek:
test this please

pgesek:
Needs an update with latest master

sms/src/main/java/org/motechproject/sms/web/StatusController.java
```diff
@@ -48,6 +47,9 @@
     private static final int RECORD_FIND_RETRY_COUNT = 3;
     private static final int RECORD_FIND_TIMEOUT = 500;
     private static final String SMS_MODULE = "motech-sms";
+    private static final String DELIVERY_CONFIRMED = "DELIVERY_CONFIRMED";
```
pgesek: we should place these in some constants class

sms/src/main/java/org/motechproject/sms/configs/Config.java
```diff
@@ -152,17 +152,17 @@ public void setProps(List<ConfigProp> props) {
...
-     * the maximum number of retries for this configuration, then {@link DeliveryStatus#ABORTED} is returned. Otherwise
-     * {@link DeliveryStatus#RETRYING} is returned, indicating the retries are ongoing.
+     * Returns an appropriate Delivery Status for the given failure count. If the failure count is higher than
+     * the maximum number of retries for this configuration, then ABORTED Status is returned. Otherwise
+     * RETRYING Status is returned, indicating the retries are ongoing.
      * @param failureCount the failure count to check
      * @return the delivery status matching the failure count
      */
-    public DeliveryStatus retryOrAbortStatus(Integer failureCount) {
+    public String retryOrAbortStatus(Integer failureCount) {
```
pgesek, line 160: Are these methods still required?

sms/src/main/java/org/motechproject/sms/http/GenericResponseHandler.java
```diff
@@ -39,9 +39,9 @@ public void handle(OutgoingSms sms, String response, Header[] headers) {
                     sms.getRecipients().toString()));
             for (String recipient : sms.getRecipients()) {
                 getAuditRecords().add(new SmsRecord(getConfig().getName(), OUTBOUND, recipient, sms.getMessage(), now(),
-                        DeliveryStatus.DISPATCHED, null, sms.getMotechId(), providerMessageId, null));
+                        DISPATCHED, null, sms.getMotechId(), providerMessageId, null));
```
pgesek: shouldn't we be placing the status from the provider here?

sms/src/main/java/org/motechproject/sms/http/MultilineSingleResponseHandler.java
```diff
@@ -50,7 +51,7 @@ public void handle(OutgoingSms sms, String response, Header[] headers) {
             getLogger().info(String.format("Sent messageId %s '%s' to %s", messageId, messageForLog(sms),
                     sms.getRecipients().get(0)));
             getAuditRecords().add(new SmsRecord(getConfig().getName(), OUTBOUND, sms.getRecipients().get(0),
-                    sms.getMessage(), now(), DeliveryStatus.DISPATCHED, null, sms.getMotechId(), messageId, null));
+                    sms.getMessage(), now(), DISPATCHED, null, sms.getMotechId(), messageId, null));
```
pgesek: shouldn't the provider status go here?

pgesek:
Looking through the PR, two issues:

1) Why do we have statuses declared as constants? Isn't the whole point of this ticket to store the status coming from the provider, not a predefined status?
2) We should consider a migration script. It should be fairly easy and allow migration in this case. Or are we storing enums as strings (I forgot)?

matkwiatkowski:
Delivery status was stored as normal string in data base so there is no need to write migration script

sms/src/main/java/org/motechproject/sms/http/MultilineResponseHandler.java
```diff
@@ -77,7 +77,7 @@ public void handle(OutgoingSms sms, String response, Header[] headers) {
                 getLogger().info(String.format("Sent messageId %s '%s' to %s", messageId, messageForLog(sms),
                         recipient));
                 getAuditRecords().add(new SmsRecord(getConfig().getName(), OUTBOUND, recipient, sms.getMessage(), now(),
-                        DeliveryStatus.DISPATCHED, null, sms.getMotechId(), messageId, null));
+                        getTemplateOutgoingResponse().getSuccessStatus(), null, sms.getMotechId(), messageId, null));
```
sebbrudzinski: According to the javadocs, this field contains the status, that the provider considers as success. Are you sure this is the proper field to use? Won't this always put the success in the logs, no matter what the actual response was?

sms/src/main/java/org/motechproject/sms/http/GenericResponseHandler.java
```diff
@@ -39,7 +38,7 @@ public void handle(OutgoingSms sms, String response, Header[] headers) {
                     sms.getRecipients().toString()));
             for (String recipient : sms.getRecipients()) {
                 getAuditRecords().add(new SmsRecord(getConfig().getName(), OUTBOUND, recipient, sms.getMessage(), now(),
-                        DeliveryStatus.DISPATCHED, null, sms.getMotechId(), providerMessageId, null));
+                        getTemplateOutgoingResponse().getSuccessStatus(), null, sms.getMotechId(), providerMessageId, null));
```
sebbrudzinski: According to the javadocs, this field contains the status, that the provider considers as success. Are you sure this is the proper field to use? Won't this always put the success in the logs, no matter what the actual response was?

sms/src/main/java/org/motechproject/sms/configs/Config.java
sebbrudzinski, line 160: What about this comment? Do we need those methods? If so, why?

sms/src/main/java/org/motechproject/sms/http/MultilineSingleResponseHandler.java
```diff
@@ -50,7 +49,7 @@ public void handle(OutgoingSms sms, String response, Header[] headers) {
             getLogger().info(String.format("Sent messageId %s '%s' to %s", messageId, messageForLog(sms),
                     sms.getRecipients().get(0)));
             getAuditRecords().add(new SmsRecord(getConfig().getName(), OUTBOUND, sms.getRecipients().get(0),
-                    sms.getMessage(), now(), DeliveryStatus.DISPATCHED, null, sms.getMotechId(), messageId, null));
+                    sms.getMessage(), now(), getTemplateOutgoingResponse().getSuccessStatus(), null, sms.getMotechId(), messageId, null));
```
sebbrudzinski: According to the javadocs, this field contains the status, that the provider considers as success. Are you sure this is the proper field to use? Won't this always put the success in the logs, no matter what the actual response was?

matkwiatkowski: Actually I think it is ok to use this status here, because in this this block we deal only with successfully send messages (it is checked by if instruction above) and in this place we put into database sms records only with DISPATCHED status.

sms/src/main/java/org/motechproject/sms/http/GenericResponseHandler.java
matkwiatkowski: Actually I think it is ok to use this status here, because in this this block we deal only with successfully send messages (it is checked by if instruction above) and in this place we put into database sms records only with DISPATCHED status.

sms/src/main/java/org/motechproject/sms/http/MultilineResponseHandler.java
matkwiatkowski: Actually I think it is ok to use this status here, because in this this block we deal only with successfully send messages (it is checked by the if instruction above) and in this place we put into database sms records only with DISPATCHED status.

sms/src/main/java/org/motechproject/sms/configs/Config.java
matkwiatkowski, line 160: Those methods are checking the number of failed attempts to sent the sms and returning the proper status (retry or abort) in order to save sms record with this status in database and resend sms when status is retry. The provider response does not have proper failure status and we still need retry mechanims so I don't see purpose in removing those methods

sms/src/main/java/org/motechproject/sms/audit/SmsRecordsDataService.java
```diff
@@ -61,7 +61,6 @@
      * @param phoneNumber the number of the phone the message was received from or delivered to
      * @param messageContent the contents of the SMS message
      * @param timestamp the date-time range the timestamp of the SMS should fall into
-     * @param deliveryStatuses the set of delivery status for the messages
```
pgesek, line 64: the param is still here

sms/src/main/java/org/motechproject/sms/configs/Config.java
pgesek, line 160: this can stay I guess

sms/src/main/java/org/motechproject/sms/http/GenericResponseHandler.java
pgesek: This is not the correct value, success status is a pattern against the response code is matched. So it takes values like '200' or '20*', it's not something we would store as the status.

sms/src/main/java/org/motechproject/sms/http/MultilineResponseHandler.java
pgesek: same comment as before

sms/src/main/java/org/motechproject/sms/http/MultilineSingleResponseHandler.java
pgesek: same comment as before

sms/src/main/java/org/motechproject/sms/service/SmsServiceImpl.java
```diff
@@ -37,6 +36,9 @@
 
     private static final Logger LOGGER = LoggerFactory.getLogger(SmsServiceImpl.class);
 
+    public static final String SCHEDULED = "SCHEDULED";
```
pgesek: why are these not in DeliveryStatus?

sms/src/main/java/org/motechproject/sms/templates/Response.java
```diff
@@ -99,6 +99,10 @@ public Boolean hasSuccessResponse() {
         return isNotBlank(successResponse);
     }
 
+    public String getSuccessStatus() {
```
pgesek: don't need this

sms/src/main/java/org/motechproject/sms/web/IncomingController.java
```diff
@@ -124,7 +125,7 @@ public void handleIncoming(@PathVariable String configName, @RequestParam Map<St
 
         eventRelay.sendEventMessage(inboundEvent(config.getName(), sender, recipient, message, providerMessageId,
                 timestamp));
-        smsRecordsDataService.create(new SmsRecord(config.getName(), INBOUND, sender, message, now(), DeliveryStatus.RECEIVED,
+        smsRecordsDataService.create(new SmsRecord(config.getName(), INBOUND, sender, message, now(), DeliveryStatuses.RECEIVED,
```
pgesek: we should use the real status from the provider.

sms/src/main/java/org/motechproject/sms/web/StatusController.java
```diff
@@ -194,14 +194,14 @@ private void analyzeStatus(Status status, String configName, Map<String, String>
...
+                smsRecord.setDeliveryStatus(statusString);
                 eventSubject = SmsEventSubjects.DELIVERY_CONFIRMED;
             } else if (status.hasStatusFailure() && statusString.matches(status.getStatusFailure())) {
-                smsRecord.setDeliveryStatus(DeliveryStatus.FAILURE_CONFIRMED);
+                smsRecord.setDeliveryStatus(statusString);
                 eventSubject = SmsEventSubjects.FAILURE_CONFIRMED;
             } else {
                 // If we're not certain the message was delivered or failed, then it's in the DISPATCHED gray area
-                smsRecord.setDeliveryStatus(DeliveryStatus.DISPATCHED);
+                smsRecord.setDeliveryStatus(DeliveryStatuses.DISPATCHED);
```
pgesek: use real status

sms/src/main/java/org/motechproject/sms/web/StatusController.java
```diff
@@ -212,7 +212,7 @@ private void analyzeStatus(Status status, String configName, Map<String, String>
                     configName, params);
             LOGGER.error(msg);
             statusMessageService.warn(msg, SMS_MODULE);
-            smsRecord.setDeliveryStatus(DeliveryStatus.FAILURE_CONFIRMED);
+            smsRecord.setDeliveryStatus(DeliveryStatuses.FAILURE_CONFIRMED);
```
pgesek, line 215: use real status

sms/src/main/java/org/motechproject/sms/http/GenericResponseHandler.java
pgesek: use dispatched here

sms/src/main/java/org/motechproject/sms/web/IncomingController.java
pgesek: you can use template.status and params for retrieving the status. Use the constant if the status.hasStatusKey returns false

sms/src/main/java/org/motechproject/sms/web/StatusController.java
matkwiatkowski, line 215: string status is null here, should we use constant or status.getStatusFailure() value?

pgesek, line 215: ok, we can leave the failure_confirmed status here

sms/src/main/java/org/motechproject/sms/web/IncomingController.java
```diff
@@ -88,28 +83,52 @@ public void handleIncoming(@PathVariable String configName, @RequestParam Map<St
         }
         Template template = templateService.getTemplate(config.getTemplateName());
 
+        eventRelay.sendEventMessage(inboundEvent(config.getName(),
+                getSenderKey(params, template),
```
pgesek: the naming is wrong, these do not return keys but actual values, they Key part should get removed

sms/src/main/java/org/motechproject/sms/web/IncomingController.java
```diff
@@ -88,28 +83,52 @@ public void handleIncoming(@PathVariable String configName, @RequestParam Map<St
...
-        }
+    private String getMessageKey(Map<String, String> params, Template template) {
+        return params.containsKey(template.getIncoming().getMessageKey()) ? params.get(template.getIncoming().getMessageKey()) : null;
+    }
 
-        if (params.containsKey(template.getIncoming().getMsgIdKey())) {
-            providerMessageId = params.get(template.getIncoming().getMsgIdKey());
-        }
+    private String getMsgIdKey(Map<String, String> params, Template template) {
+        return params.containsKey(template.getIncoming().getMsgIdKey()) ? params.get(template.getIncoming().getMsgIdKey()) : null;
```
pgesek: the checks are redundant, map will return a null if it doesn't contain a key

sms/src/main/java/org/motechproject/sms/web/IncomingController.java
```diff
@@ -117,14 +136,12 @@ public void handleIncoming(@PathVariable String configName, @RequestParam Map<St
...
         }
+        return now();
+    }
 
-        eventRelay.sendEventMessage(inboundEvent(config.getName(), sender, recipient, message, providerMessageId,
-                timestamp));
-        smsRecordsDataService.create(new SmsRecord(config.getName(), INBOUND, sender, message, now(), DeliveryStatus.RECEIVED,
-                null, null, providerMessageId, null));
+    private String getStatusKey(Map<String, String> params, Template template) {
+        return template.getStatus().hasStatusKey() && params.containsKey(template.getStatus().getStatusKey()) ? template.getStatus().getStatusKey() : DeliveryStatuses.RECEIVED;
```
pgesek: we shouldn't be saving the key, but the actual status from the params.

kmadej:
test this please

kmadej:
test this please

PR #131, opened 2016-02-10
MOTECH-2172: Fixed data forwarding rule handling

pgesek:
- Saving config won't fall apart if the user doesn't have access to
  the forwarding config, which was the root of the error
- If forwarding rules fail to change, this will be reflected on the UI
  (albeit this can still use some improvement)
- Fixed a js bug with failure handling in the verify method (leftover
  code)


sebbrudzinski:
@pgesek OK to merge

PR #132, opened 2016-02-10
MOTECH-2161: Adds a task data provider to OpenMRS module

kmadej:
This is first PR in this issue. It adds a task data provider to OpenMRS module.
Now we can get Provider object using getProviderByUuid lookup. For this
data source we've got the following information available:

Provider UUID
Provider identifier
Person UUID
Person display

I had to add display field to OpenMRSPerson, which is our MOTECH representation
of Person.

Next lookups like getPatientByUuid, getPatientByMotechId, getEncounterByUuid
will be added in next PR


openmrs-19/src/main/java/org/motechproject/openmrs19/tasks/OpenMRSTaskDataProvider.java
```diff
@@ -0,0 +1,86 @@
...
+import static org.motechproject.openmrs19.tasks.OpenMRSTasksConstants.UUID;
+
+/**
+ * This is the OpenMRS task data provider that is registered with the task module as a data source.
+ * It allows retrieving objects from OpenMRS model and using it in tasks.
+ */
+@Service("openMRSTaskDataProvider")
+public class OpenMRSTaskDataProvider extends AbstractDataProvider {
+    private static final Logger LOGGER = LoggerFactory.getLogger(OpenMRSTaskDataProvider.class);
+    private static final List<Class<?>> SUPPORTED_CLASS = Arrays.asList(OpenMRSProvider.class);
```
frankhuster: SUPPORTED_CLASSES instead? I Imagine there'll be more than one element soon, right?

openmrs-19/src/main/resources/webapp/messages/messages_es.properties
```diff
@@ -0,0 +1,20 @@
+#Module name
```
frankhuster: I thought you only needed to add these files if they were different, can you confirm?

openmrs-19/src/main/resources/webapp/messages/messages_fr.properties
```diff
@@ -0,0 +1,20 @@
+#Module name
```
frankhuster: I thought you only needed to add these files if they were different, can you confirm?

openmrs-19/src/main/java/org/motechproject/openmrs19/tasks/OpenMRSTaskDataProvider.java
kmadej: Yes, as I wrote, there will be more supported classes soon (I have to change name of variable to SUPPORTED_CLASSES). Now there is only Provider, because I have still some questions about Encounter and Patient.

openmrs-19/src/main/resources/webapp/messages/messages_es.properties
kmadej: Probably you are right, I will change it

frankhuster:
Couple of nitpicks that maybe you can fix with the next PR?

LGTM

openmrs-19/src/main/resources/webapp/messages/messages_es.properties
kmadej: I checked it, and now for sure you are right

kmadej:
I will push this small changes in this PR

PR #133, opened 2016-02-10
MOTECH-2171: (CC) Made disabled button actually disabled

pgesek:
ng-disabled won't disable ng-click on < a > or < div > it
seems, I used a logical and to make sure it won't work.

### DIFF
--- a/commcare/src/main/resources/webapp/partials/settings.html
+++ b/commcare/src/main/resources/webapp/partials/settings.html
@@ -17,10 +17,10 @@
              <a ng-if="selectedConfig" ng-click="deleteConfig()" type="button" class="btn btn-danger input-small">
                  <i class="fa fa-fw fa-trash-o"></i> {{msg('commcare.delete')}}
              </a>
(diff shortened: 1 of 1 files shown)


commcare/src/main/resources/webapp/partials/settings.html
```diff
@@ -11,7 +11,7 @@
              <a ng-click="addConfig()" type="button" class="form-group btn btn-success">
                  <i class="fa fa-plus"></i> {{msg('commcare.addConfig')}}
              </a>
-             <a ng-if="selectedConfig" ng-disabled="!saveAllowed()" ng-click="saveConfig()" type="button" class="btn btn-primary">
+             <a ng-if="selectedConfig" ng-disabled="!saveAllowed()" ng-click="saveAllowed() && saveConfig()" type="button" class="btn btn-primary">
```
sebbrudzinski, line 14: If ng-disabled does not work on anchor links, we should either change them to buttons, or adjust the ng-click logic in the remaining buttons as well.

sebbrudzinski:
@pgesek OK to merge

PR #136, opened 2016-02-11
MOTECH-2161: Adds getEncounterByUuid lookup

kmadej:
This PR adds next lookup to a task data source in OpenMRS module. I had to
adjust OpenMRS model in our Motech implementation. The following changes
had to be done:
- add display field to Encounter and OpenMRSEncounter
- add display field to Location and OpenMRSFacility

OpenMRSEncounter and OpenMRSFacility are Motech representation of Encounter
and Location model.

After this changes we have getEncounterByUuid lookup in OpenMRS task data source,
with the following fields :

Encounter UUID
Encounter display
Encounter date
Provider UUID
Provider display
Location UUID
Location display
Patient UUID
Patient display

In the spec for this issue there is information about OpenMRS Form, which is
related with Encounter. However, now we don't have Motech representation of
OpenMRS Form and that's why this field is not included in Encounter fields.
I will do it in another PR.

Lookups getPatientByUuid and getPatientByMotechId will be added in the separate
PR.

### DIFF
--- a/openmrs-19/src/main/resources/task-data-provider.json
+++ b/openmrs-19/src/main/resources/task-data-provider.json
@@ -14,7 +14,7 @@
       ],
       "fields": [
         {
-          "displayName": "openMRS.provider.providerId",
+          "displayName": "openMRS.provider.id",
           "fieldKey": "providerId"
         },
         {
@@ -28,6 +28,61 @@
         {
           "displayName": "openMRS.person.display",
           "fieldKey": "person.display"
+        }
+      ]
+    },
+    {
+      "displayName": "openMRS.encounter",
+      "type": "OpenMRSEncounter",
+      "lookupFields": [
+        {
+          "displayName": "openMRS.lookup.uuid",
+          "fields": [
+            "openMRS.uuid"
+          ]
+        }
+      ],
+      "fields": [
+        {
+          "displayName": "openMRS.encounter.id",
+          "fieldKey": "id"
+        },
+        {
+          "displayName": "openMRS.encounter.display",
+          "fieldKey": "display"
+        },
+        {
+          "displayName": "openMRS.encounter.date",
+          "fieldKey": "date",
+          "type": "DATE"
+        },
+        {
+          "displayName": "openMRS.encounter.type",
+          "fieldKey": "encounterType"
+        },
+        {
+          "displayName": "openMRS.provider.id",
+          "fieldKey": "provider.providerId"
+        },
+        {
+          "displayName": "openMRS.provider.display",
+          "fieldKey": "provider.person.display"
+        },
+        {
+          "displayName": "openMRS.location.id",
+          "fieldKey": "facility.id"
+        },
+        {
+          "displayName": "openMRS.location.display",
+          "fieldKey": "facility.display"
+        },
+        {
+          "displayName": "openMRS.patient.id",
+          "fieldKey": "patient.id"
+        },
+        {
+          "displayName": "openMRS.patient.display",
+          "fieldKey": "patient.person.display"
         }
       ]
     }
--- a/openmrs-19/src/test/java/org/motechproject/openmrs19/tasks/OpenMRSTaskDataProviderTest.java
+++ b/openmrs-19/src/test/java/org/motechproject/openmrs19/tasks/OpenMRSTaskDataProviderTest.java
@@ -5,7 +5,9 @@ import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.mockito.Mock;
 import org.mockito.runners.MockitoJUnitRunner;
+import org.motechproject.openmrs19.domain.OpenMRSEncounter;
 import org.motechproject.openmrs19.domain.OpenMRSProvider;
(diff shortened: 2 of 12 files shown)


openmrs-19/src/main/resources/task-data-provider.json
```diff
@@ -14,7 +14,7 @@
       ],
       "fields": [
         {
-          "displayName": "openMRS.provider.providerId",
+          "displayName": "openMRS.provider.id",
           "fieldKey": "providerId"
```
frankhuster, line 18: Sometimes the fieldKey value is in the form "person.id" and sometimes it's in the form "providerId", do you think we should make it consistent (either camelCase or period.delimited)?

openmrs-19/src/test/java/org/motechproject/openmrs19/tasks/OpenMRSTaskDataProviderTest.java
```diff
@@ -47,6 +52,57 @@ public void shouldReturnNullWhenClassIsNotSupported() {
...
+
+        when(encounterService.getEncounterByUuid("4")).thenReturn(null);
+        Object object = taskDataProvider.lookup(className, BY_UUID, lookupFields);
+
+        assertNull(object);
+        verify(encounterService).getEncounterByUuid("4");
+    }
+
+    @Test
+    public void shouldReturnEncounterForLookupGetEncounterByUuid() {
```
frankhuster, line 92: Is that test mostly testing when().thenReturn() or am I missing something?

openmrs-19/src/main/resources/task-data-provider.json
kmadej, line 18: fieldKey depends on the name of field in returning object. For example OpenMRSProvider has field "providerId" and OpenMRSPerson "id". FieldKeys are connected with Java classes variable names.  So we have to change variable names in our OpenMRS model and then we can change fieldKeys. 

The value of displayName depends on us and it is connected with messages.properties

openmrs-19/src/test/java/org/motechproject/openmrs19/tasks/OpenMRSTaskDataProviderTest.java
kmadej, line 92: The most important thing in this test is assert checking if object is null and verification if getEncounterByUuid method was invoked with "4".

when().thenReturn() is faking behaviour of encounterService. A user is trying to retrieve Encounter with Uuid "4" and we want to return null, which means that Encounter is not found. However in the test we can define many when().thenReturn(), so at the end of the test we make sure that getEncounterByUuid was invoked with "4"

kmadej, line 92: Additionally the default behaviour for mock objects is 'return null if when not defined', so in this test when().thenReturn() is redundant, because even if we not defined this the encounterService mock will return null. However in my opinion it is more clear when reading the code.

openmrs-19/src/test/java/org/motechproject/openmrs19/tasks/OpenMRSTaskDataProviderTest.java
```diff
@@ -47,6 +52,57 @@ public void shouldReturnNullWhenClassIsNotSupported() {
...
+        Map<String, String> lookupFields = new HashMap<>();
+        lookupFields.put(UUID, "5");
+
+        OpenMRSEncounter openMRSEncounter = new OpenMRSEncounter();
+        openMRSEncounter.setEncounterType("encounterTypeTest");
+
+        when(encounterService.getEncounterByUuid("5")).thenReturn(openMRSEncounter);
+        Object object = taskDataProvider.lookup(className, BY_UUID, lookupFields);
+
+        assertTrue(object instanceof OpenMRSEncounter);
```
pgesek: shouldn't we check that object equals to openMRSEncounter ? It should be the same instance

openmrs-19/src/test/java/org/motechproject/openmrs19/tasks/OpenMRSTaskDataProviderTest.java
```diff
@@ -47,6 +52,57 @@ public void shouldReturnNullWhenClassIsNotSupported() {
...
+    public void shouldReturnEncounterForLookupGetEncounterByUuid() {
+        String className = OpenMRSEncounter.class.getSimpleName();
+        Map<String, String> lookupFields = new HashMap<>();
+        lookupFields.put(UUID, "5");
+
+        OpenMRSEncounter openMRSEncounter = new OpenMRSEncounter();
+        openMRSEncounter.setEncounterType("encounterTypeTest");
+
+        when(encounterService.getEncounterByUuid("5")).thenReturn(openMRSEncounter);
+        Object object = taskDataProvider.lookup(className, BY_UUID, lookupFields);
```
pgesek, line 102: since this this is the 'then' line, can we separate it from the line above?

kmadej:
test this please

kmadej:
test this please

PR #137, opened 2016-02-12
MOTECH-2170: Fixed showing ownerId in Commcare Cases

pkornowski:
Owner ID in Commcare Cases was displayed as null. That was caused by wrong gained parameter.

### DIFF
--- a/commcare/src/main/java/org/motechproject/commcare/service/impl/CommcareCaseServiceImpl.java
+++ b/commcare/src/main/java/org/motechproject/commcare/service/impl/CommcareCaseServiceImpl.java
@@ -80,6 +80,16 @@ public class CommcareCaseServiceImpl implements CommcareCaseService {
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


commcare/src/main/java/org/motechproject/commcare/service/impl/CommcareCaseServiceImpl.java
```diff
@@ -280,7 +280,7 @@ private CaseInfo populateCaseInfo(CaseJson caseResponse, String configName) {
 
         String caseType = properties.get("case_type");
         String dateOpened = properties.get("date_opened");
-        String ownerId = properties.get("owner_id)");
+        String ownerId = properties.get("owner_id");
```
pgesek, line 298: can we extend CommcareCaseServiceImplTest to test fr this field, not only for CaseId as it does currently?

PR #142, opened 2016-02-15
MOTECH-2161: Adds getPatientByUuid and getPatientByMotechId lookups

kmadej:
This commit adds two lookups to a task data source in OpenMRS module : getPatientByUuid and getPatientByMotechId.

The object fields for Patient are :
- Patient UUID
- Patient identifiers
- Motech ID
- Person UUID
- Person Display
- Person address
- Person age
- Person birthDateEstimated flag
- Person date of birth
- Person date of death
- Person gender
- Person dead flag

I had to make some changes in OpenMRS model. Previously we supported only Motech ID as an identifier for Patient. Now we can get others identifiers from Patient. To define identifiers, which we want to retrieve from Patient, we should add a property to openmrs.properties. For example :
- openmrs.identifierTypes=CommCare CaseID.

Then, we will get the value for mentioned identifiers. 

Additionally I had to change getter name for dead field in OpenMRSPerson. A task data source doesn't properly read value if the getter for boolean has 'isDead' form.

Moreover I changed validation for OpenMRSPatient before save. Now OpenMRS doesn't require a location for Patient identifier.  

Also I added few tests.

### DIFF
--- a/openmrs-19/src/main/java/org/motechproject/openmrs19/OpenMrsInstance.java
+++ b/openmrs-19/src/main/java/org/motechproject/openmrs19/OpenMrsInstance.java
@@ -1,5 +1,6 @@
 package org.motechproject.openmrs19;
 
+import org.apache.commons.lang.StringUtils;
 import org.motechproject.commons.api.MotechException;
 import org.motechproject.server.config.SettingsFacade;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -10,6 +11,9 @@ import org.springframework.web.util.UriTemplate;
 import javax.annotation.PostConstruct;
 import java.net.URI;
 import java.net.URISyntaxException;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
 
 /**
  * Represents a single OpenMRS Web Application instance.
@@ -17,12 +21,14 @@ import java.net.URISyntaxException;
 @Component
 public class OpenMrsInstance {
 
+    private static final String OPENMRS_IDENTIFIER_TYPES_PROPERTY = "openmrs.identifierTypes";
     private static final String OPENMRS_MOTECH_ID_NAME_PROPERTY = "openmrs.motechIdName";
     private static final String OPENMRS_URL_PROPERTY = "openmrs.url";
     private static final String OPENMRS_WEB_SERVICE_PATH = "/ws/rest/v1";
 
     private String openmrsUrl;
     private String motechPatientIdentifierTypeName;
+    private List<String> patientIdentifierTypesNames;
 
     private SettingsFacade settingsFacade;
 
@@ -35,6 +41,7 @@ public class OpenMrsInstance {
     public void readSettings() {
         this.openmrsUrl = settingsFacade.getProperty(OPENMRS_URL_PROPERTY) + OPENMRS_WEB_SERVICE_PATH;
         this.motechPatientIdentifierTypeName = settingsFacade.getProperty(OPENMRS_MOTECH_ID_NAME_PROPERTY);
+        this.patientIdentifierTypesNames = parseIdentifierTypesProperty();
     }
 
     /**
@@ -69,5 +76,15 @@ public class OpenMrsInstance {
 
     public String getMotechPatientIdentifierTypeName() {
         return motechPatientIdentifierTypeName;
+    }
+
+    public List<String> getPatientIdentifierTypesNames() {
+        return patientIdentifierTypesNames;
+    }
+
+    private List<String> parseIdentifierTypesProperty() {
+        String property = settingsFacade.getProperty(OPENMRS_IDENTIFIER_TYPES_PROPERTY);
+
+        return StringUtils.isEmpty(property) ? new ArrayList<>() : Arrays.asList(property.split(","));
     }
 }
--- a/openmrs-19/src/main/java/org/motechproject/openmrs19/OpenMrsInstance.java
+++ b/openmrs-19/src/main/java/org/motechproject/openmrs19/OpenMrsInstance.java
@@ -28,7 +28,7 @@ public class OpenMrsInstance {
 
     private String openmrsUrl;
     private String motechPatientIdentifierTypeName;
-    private List<String> patientIdentifierTypesNames;
+    private List<String> patientIdentifierTypeNames;
 
     private SettingsFacade settingsFacade;
 
@@ -41,7 +41,7 @@ public class OpenMrsInstance {
     public void readSettings() {
         this.openmrsUrl = settingsFacade.getProperty(OPENMRS_URL_PROPERTY) + OPENMRS_WEB_SERVICE_PATH;
         this.motechPatientIdentifierTypeName = settingsFacade.getProperty(OPENMRS_MOTECH_ID_NAME_PROPERTY);
-        this.patientIdentifierTypesNames = parseIdentifierTypesProperty();
+        this.patientIdentifierTypeNames = parseIdentifierTypesProperty();
     }
 
     /**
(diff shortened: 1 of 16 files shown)


openmrs-19/src/main/java/org/motechproject/openmrs19/domain/OpenMRSPatient.java
```diff
@@ -34,23 +36,39 @@ public OpenMRSPatient(String patientId) {
...
+     * Creates a patient with the given {@code motechId} and {@code identifierTypesByValue} based on the given {@code person}
      * details and assigns it to the given {@code facility}.
      *
+     * @param motechId  the MOTECH ID of the patient
+     * @param person  the personal details about the patient
+     * @param facility  the facility by which the patient is treated
+     * @param identifierTypesByValue the identifiers of patient by value
+     */
+    public OpenMRSPatient(String motechId, OpenMRSPerson person, OpenMRSFacility facility, Map<String, String> identifierTypesByValue) {
+        this(null, motechId, person, facility, identifierTypesByValue);
```
sebbrudzinski: ByValue? What value? Can you just name it identifiers and explain in the javadoc what is the key and what is the value?

openmrs-19/src/main/java/org/motechproject/openmrs19/OpenMrsInstance.java
```diff
@@ -17,12 +21,14 @@
...
 public class OpenMrsInstance {
 
+    private static final String OPENMRS_IDENTIFIER_TYPES_PROPERTY = "openmrs.identifierTypes";
     private static final String OPENMRS_MOTECH_ID_NAME_PROPERTY = "openmrs.motechIdName";
     private static final String OPENMRS_URL_PROPERTY = "openmrs.url";
     private static final String OPENMRS_WEB_SERVICE_PATH = "/ws/rest/v1";
 
     private String openmrsUrl;
     private String motechPatientIdentifierTypeName;
+    private List<String> patientIdentifierTypesNames;
```
sebbrudzinski: Types or TypeNames

openmrs-19/src/main/java/org/motechproject/openmrs19/domain/OpenMRSPerson.java
```diff
@@ -161,8 +161,9 @@ public Boolean getBirthDateEstimated() {
         return birthDateEstimated;
     }
 
-    public Boolean isDead() {
-        return isDead;
+    // with getter isDead() task data source doesn't properly read this value
```
sebbrudzinski: Is it a problem in the Tasks module? If so, file a bug, append "TODO", and provide ticket ID.

openmrs-19/src/main/java/org/motechproject/openmrs19/resource/PatientResource.java
```diff
@@ -39,12 +39,21 @@
...
      * Returns the UUID of the MOTECH patient identifier.
      *
-     * @return  the UUD of the MOTECH patient identifier
+     * @return  the UUID of the MOTECH patient identifier
      * @throws HttpException  when there were problems while fetching MOTECH patient identifier
      */
     String getMotechPatientIdentifierUuid() throws HttpException;
 
     /**
+     * Returns the patient identifier name for the given uuid.
```
sebbrudzinski: This seems to perform various other operations than just returning patient identifier name. Please correct the javadoc to include everything that happens after calling this method and preferably rename this method, as it's not just a "get". Moreover, it's not clear to me what this method returns. Correct that as well.

openmrs-19/src/main/java/org/motechproject/openmrs19/resource/impl/PatientResourceImpl.java
```diff
@@ -83,6 +87,27 @@ public String getMotechPatientIdentifierUuid() throws HttpException {
...
     @Override
+    public String getPatientIdentifierName(String uuid) throws HttpException {
+        String identifierName = identifierUuidByName.get(uuid);
+
+        if (identifierName == null) {
+            PatientIdentifierListResult result = getAllPatientIdentifierTypes();
+            for (IdentifierType type : result.getResults()) {
+                if (StringUtils.equals(uuid, type.getUuid())) {
+                    if (openmrsInstance.getPatientIdentifierTypesNames().contains(type.getName())) {
+                        identifierName = type.getName();
```
sebbrudzinski: Is it supposed to return identifier name, based on the value of that identifier? What if I have the same value for various identifiers? Eg. MotechID=1;  RandomID=1;  OneMoreID=1?

openmrs-19/src/main/java/org/motechproject/openmrs19/service/impl/OpenMRSPatientServiceImpl.java
```diff
@@ -93,20 +94,44 @@ public OpenMRSPatient getPatientByUuid(String uuid) {
...
+            if (motechIdentifier.getLocation() != null) {
+                facility = facilityAdapter.getFacilityByUuid(motechIdentifier.getLocation().getUuid());
             }
 
-            motechId = identifier.getIdentifier();
+            motechId = motechIdentifier.getIdentifier();
         }
 
-        return ConverterUtils.toOpenMRSPatient(patient, facility, motechId);
+        List<Identifier> supportedIdentifierList = new ArrayList<>();
```
sebbrudzinski: Can the process for discovering supported identifiers be moved to separate, private method?

openmrs-19/src/main/java/org/motechproject/openmrs19/util/ConverterUtils.java
```diff
@@ -371,13 +375,21 @@ public static OpenMRSPatient toOpenMRSPatient(Patient patient) {
      * @param motechId  the MOTECH ID of the patient
      * @return
      */
-    public static OpenMRSPatient toOpenMRSPatient(Patient patient, OpenMRSFacility facility, String motechId) {
-
+    public static OpenMRSPatient toOpenMRSPatient(Patient patient, OpenMRSFacility facility, String motechId,
+                                                  List<Identifier> supportedIdentifierList) {
         OpenMRSPatient openMRSPatient = new OpenMRSPatient(patient.getUuid());
 
+        Map<String, String> identifierTypesByValue = new HashMap<>();
```
sebbrudzinski: Again, ByValue?

openmrs-19/src/main/resources/openmrs.properties
```diff
@@ -2,3 +2,4 @@ openmrs.url=http://localhost:8080/openmrs
 openmrs.user=admin
 openmrs.password=Admin123
 openmrs.motechIdName=MOTECH Id
+openmrs.identifierTypes=CommCare CaseID
```
sebbrudzinski: Should probably be empty by default, since this is implementation-specific

openmrs-19/src/main/java/org/motechproject/openmrs19/resource/impl/PatientResourceImpl.java
kmadej: I know that it is complicated and it's not easy to understand what happens in this module. I will try to explain you how it works and maybe you can suggest how to improve understandability of this code. Personally, I assume that we should write a documentation for OpenMRS.
1. In OpenMRS we can define Patient Identifier Types, for example : MOTECH ID, CommCare CaseID, OpenMRS ID. 
2. Every Identifier Type has unique name and uuid, which is generated by OpenMRS.
3. When we create Patient, we can add Identifiers giving name of the Identifier Type and identifier number.

In this method we retrieve the Patient Identifier Type name (OpenMRS ID, CommCare CaseID etc.) giving Uuid.

After we have the name of Identifier Type, we can add an Identifier to Patient.  Identifier object looks like the following :
- String identifier (value, number for Patient identifier, for example 100251)
- IdentifierType identifierType (object with identifier name, for example : MOTECH ID)

So answering to your question, we don't retrieve identifier type name by identifier value, but by their uuid. Here there is no connection with concrete Patient object.

I hope, that it is clear to you. Maybe I can write this method better or extend javadoc. Let me know what do you think about it.

sebbrudzinski: Thanks, it's clear. It can stay like this, just update the javadoc to reflect what this method really does.

PR #147, opened 2016-02-17
MOTECH-2162: Adds a task action createPatient to OpenMRS module

kmadej:
This PR creates a task action createPatient for OpenMRS module.
This action has the following fields:
- String firstName (required)
- String middleName
  -String lastName (required)
- String address
- Date dateOfBirth
- Boolean birthDateEstimated
- String gender (required)
- Boolean dead
- String motechId (required)
- String locationName
- Map identifiers (key - identifier type name, value - identifier number)

If the locatioName is not provided, the default location will be used for
identifiers.

To store identifier to patient, it is required to define identifier type
name in property openmrs.identifierTypes in openmrs.properties file.
Then you can add a pair to identifiers map. This feature needs https://github.com/motech/modules/pull/145

### DIFF
--- a/openmrs-19/src/main/java/org/motechproject/openmrs19/tasks/impl/OpenMRSActionProxyServiceImpl.java
+++ b/openmrs-19/src/main/java/org/motechproject/openmrs19/tasks/impl/OpenMRSActionProxyServiceImpl.java
@@ -1,0 +1,80 @@
+package org.motechproject.openmrs19.tasks.impl;
+
+import org.apache.commons.lang.StringUtils;
+import org.joda.time.DateTime;
+import org.motechproject.openmrs19.domain.OpenMRSFacility;
+import org.motechproject.openmrs19.domain.OpenMRSPatient;
+import org.motechproject.openmrs19.domain.OpenMRSPerson;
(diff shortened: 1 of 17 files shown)


openmrs-19/src/main/java/org/motechproject/openmrs19/tasks/impl/OpenMRSActionProxyServiceImpl.java
```diff
@@ -0,0 +1,80 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Service;
+
+import java.util.List;
+import java.util.Map;
+
+@Service("openMRSActionProxyService")
+public class OpenMRSActionProxyServiceImpl implements OpenMRSActionProxyService {
```
pgesek, line 23: javadoc

pgesek, line 23: add unit tests

PR #148, opened 2016-02-17
MOTECH-2148 Added ability to upload forms to CC via tasks action

sebbrudzinski:
A new task action is now generated for each form present
in the Commcare. If form question has got predefined values,
they are generated as dropdown options.

### DIFF
--- a/commcare/src/main/java/org/motechproject/commcare/events/FormEventHandler.java
+++ b/commcare/src/main/java/org/motechproject/commcare/events/FormEventHandler.java
@@ -1,0 +1,83 @@
+package org.motechproject.commcare.events;
+
+import org.apache.commons.lang.StringUtils;
+import org.joda.time.DateTime;
+import org.joda.time.format.DateTimeFormat;
+import org.joda.time.format.DateTimeFormatter;
+import org.motechproject.commcare.domain.FormValueElement;
+import org.motechproject.commcare.domain.FormXml;
+import org.motechproject.commcare.domain.MetadataValue;
+import org.motechproject.commcare.events.constants.EventSubjects;
+import org.motechproject.commcare.service.CommcareFormService;
+import org.motechproject.event.MotechEvent;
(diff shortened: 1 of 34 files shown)


commcare/src/main/java/org/motechproject/commcare/events/FormEventHandler.java
```diff
@@ -0,0 +1,83 @@
...
+import java.util.Map;
+import java.util.UUID;
+
+/**
+ * This class serves as the event handler for the task actions, exposed by the Commcare module.
+ * Respective methods extract the necessary data from the {@link MotechEvent} instance and
+ * pass them to the {@link CommcareFormService} that handles all the operations on Commcare forms.
+ */
+@Component
+public class FormEventHandler {
```
pgesek: Rename this to FormActionEventHandler - so that the purpose is obvious.

commcare/src/main/java/org/motechproject/commcare/events/FormEventHandler.java
```diff
@@ -0,0 +1,83 @@
...
+
+    private static final DateTimeFormatter COMMCARE_DATETIME_FORMAT = DateTimeFormat.forPattern("yyyy-MM-dd'T'HH:mm:ss.SSS'Z'");
+
+    @Autowired
+    private CommcareFormService commcareFormService;
+
+    @MotechListener(subjects = EventSubjects.SEND_FORM + ".*")
+    public void sendForm(MotechEvent event) {
+        String configName = EventSubjects.getConfigName(event.getSubject());
+        String xmlns = StringUtils.removeEnd(event.getSubject().replaceFirst(EventSubjects.SEND_FORM + ".", StringUtils.EMPTY), "." + configName);
```
pgesek: I don't get it, is the config name appended at the end after the xmlns? I would expect it the other way around.

Probably a good candidate for a comment at least.

PR #151, opened 2016-02-19
MOTECH-1946: Send Form -> Submit Form

pgesek:
Also fixed that stupid double 'Forms' tab issue


commcare/src/main/java/org/motechproject/commcare/events/FormActionEventHandler.java
```diff
@@ -42,7 +42,7 @@
 
     /**
      * Handles events, connected with sending Commcare forms. The event subject should have the following syntax:
-     * {@code EventSubjects.SEND_FORM.[form XMLNS].[config name]}
+     * {@code EventSubjects.SUBMIT_FORM.[form XMLNS].[config name]}
```
sebbrudzinski, line 45: This should not be changed, as it's the event subject, unless you want to rename the event subjects too.

PR #152, opened 2016-02-20
MOTECH-2200: add 3 new properties to Person

frankhuster:
Fixes MOTECH-2200

Added dateCreated, dateChanged and newPerson fields to Patient OpenMRS entity.
As of now, newPerson returns "yes" or "no", but should be changed to bool when Tasks can filter on boolean fields.


frankhuster:
Shouldn't have created this including the Atom Client module changes, will re-create

PR #153, opened 2016-02-20
MOTECH-2200: add new properties to OpenMRS Patient

frankhuster:
Added dateCreated, dateChanged and newPerson fields to Patient OpenMRS entity. 
As of now, newPerson returns "yes" or "no", but should be changed to bool when Tasks can filter on boolean fields.


pgesek:
@frankhuster I've merged this, but you will probably also have to ge this to 0.28.X. Sorry for the inconvenience, could probably merge this before branching off.

PR #188, opened 2016-03-02
MOTECH-2202: fix failing IT test

sebastianceronik:
1. Removed additional "," wich we didnt need in FullAddressString
2. Fixed arguments order in assertEquals function
3. Since we are able to convert addresses from openMRS to Motech representation only for now, I changed assert to check if address contains String rather than is equal to. that is best way to do it for now I belive. We will have to change that in future when we will be able to support openMRS address fully.

### DIFF
--- a/openmrs-19/src/test/java/org/motechproject/openmrs19/it/MRSPersonServiceIT.java
+++ b/openmrs-19/src/test/java/org/motechproject/openmrs19/it/MRSPersonServiceIT.java
@@ -70,11 +70,6 @@ public class MRSPersonServiceIT extends BasePaxIT {
         assertFalse(mrsListener.updated);
     }
 
-    /*
-        We are able to convert fields from openMRS to single string field only for now
-        that is why we are checking assertTrue with contains function for changed address
-        rather than assertEquals for string field
-     */
     @Test
     public void shouldUpdatePerson() throws OpenMRSException, InterruptedException {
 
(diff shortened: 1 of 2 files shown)


openmrs-19/src/test/java/org/motechproject/openmrs19/it/MRSPersonServiceIT.java
```diff
@@ -61,15 +61,20 @@ public void initialize() throws InterruptedException {
...
+        assertEquals(mrsListener.eventParameters.get(EventKeys.PERSON_ID), person.getPersonId());
+        assertEquals(mrsListener.eventParameters.get(EventKeys.PERSON_FIRST_NAME), person.getFirstName());
+        assertEquals(mrsListener.eventParameters.get(EventKeys.PERSON_LAST_NAME), person.getLastName());
 
         assertTrue(mrsListener.created);
         assertFalse(mrsListener.deleted);
         assertFalse(mrsListener.updated);
     }
 
+    /*
```
kmadej: add comment about address above line with assertTrue connected with address. Do not add javadoc to test method.

Moreover this comment is unclear. I suggest the following :
'So far OpenMRS module stores only one field of person's address, which is 'address1'. However while retrieving person from OpenMRS server all person's address fields are put into one string. That's why it is checked if address field contains address1 value.'

PR #190, opened 2016-03-03
MOTECH-2202: add unit test for getting address fields

sebastianceronik:
1 Changed misspelled countyDistrict.
2 Added Unit test for checking all Adress fields we get from openmrs.

### DIFF
--- a/openmrs-19/src/test/java/org/motechproject/openmrs19/resource/impl/PersonResourceImplTest.java
+++ b/openmrs-19/src/test/java/org/motechproject/openmrs19/resource/impl/PersonResourceImplTest.java
@@ -42,43 +42,14 @@ public class PersonResourceImplTest extends AbstractResourceImplTest {
     }
 
     @Test
-    public void shouldGetAllAdressFields() throws HttpException, IOException {
+    public void shouldGetAllAddressFields() throws HttpException, IOException {
         Mockito.when(getClient().getJson(Mockito.any(URI.class)))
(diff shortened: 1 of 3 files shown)


openmrs-19/src/test/java/org/motechproject/openmrs19/resource/impl/PersonResourceImplTest.java
```diff
@@ -42,6 +42,46 @@ public void shouldGetPersonById() throws HttpException, IOException {
...
+        String address5 = null;
+        String address6 = null;
+        String startDate = "2016-03-01T00:00:00.000+0000";
+        String endDate = null;
+        String latitude = "47.613879";
+        String longitude = "-122.342436";
+
+        Person person = impl.getPersonById("PPP");
+
+        assertEquals(address1, person.getPreferredAddress().getAddress1());
```
kmadej: here you are only checking json parsing, you should rather check if method getFullAddressString in Person class is properly merging all address fields into one string.

openmrs-19/src/test/java/org/motechproject/openmrs19/resource/impl/PersonResourceImplTest.java
```diff
@@ -42,6 +42,46 @@ public void shouldGetPersonById() throws HttpException, IOException {
     }
 
     @Test
+    public void shouldGetAllAdressFields() throws HttpException, IOException {
```
kmadej: Address

PR #201, opened 2016-03-09
MOTECH-2207 Fix up button groups in UI

martokarski:
- Buttons that immedately effect the current screen or object in a row: Pull to the right
- Buttons that are secondary (such as filtering): Pull to the left
- All buttons are gray (.btn-default)
- Submit/Save/Confirm buttons are .btn-primary
- Add/create buttons are green (.btn-success)
- Destructive buttons (delete ect) are red (.btn-danger)


kmadej:
I think this button should be gray or blue. I noticed that in MOTECH imports button like 'import data', 'import csv' etc are gray, so I suggest also gray colour for this button.

screenshot from 2016-03-10 11 38 08

nickdotreid:
I'd make that button blue as its the primary action for that screen, and if
it were a form, it would be an input type=submit

martokarski:
Added confirm modal when remove saved setting in commcare. Unsaved settings have another modal with warning about lost changes.

commcare/src/main/resources/webapp/messages/messages.properties
```diff
@@ -201,6 +201,7 @@ commcare.error=Error
 commcare.error.cases=Unable to retrieve the list of cases
 commcare.error.schema=Unable to retrieve application schema
 
+commcare.confirm.removeSetting=Are you sure you want to remove this setting?
```
kmadej: I think we should stick with names from UI, so I propose : "Are you sure you want to delete this configuration"

Moreover you can change the property name to "commcare.confirm.deleteConfig"

kmadej:
test this please

PR #203, opened 2016-03-10
Fixed failing build of the message-campaign module

ngraczewski:
Requires https://github.com/motech/motech/pull/420


message-campaign/src/test/java/org/motechproject/messagecampaign/it/MessageCampaignServiceBundleIT.java
```diff
@@ -63,7 +63,7 @@ public void setup() {
     public void shouldUnscheduleMessageJobsWhenCampaignIsStopped() throws SchedulerException {
         CampaignRequest campaignRequest = new CampaignRequest("entity_1", "PREGNANCY", new LocalDate(2020, 7, 10), null);
 
-        TriggerKey triggerKey = triggerKey("org.motechproject.messagecampaign.fired-campaign-message-MessageJob.PREGNANCY.entity_1.PREGNANCY", "default");
+        TriggerKey triggerKey = triggerKey("org.motechproject.messagecampaign.fired-campaign-message-MessageJob.PREGNANCY.entity_1.PREGNANCY-dayofweek", "default");
```
sebbrudzinski, line 66: Can we also have tests for no prefix version? It's essential that campaigns created in the past still work after this change.

sebbrudzinski:
It seems like the path we have chosen is dead end and introduces new problems in more places. Let's revert the changes to day of week and cron suffixes and simply add a flag in job data to recognize the jobs for edit purposes.

PR #207, opened 2016-03-11
MOTECH-2270: Fixed issue with events by handled 2x times

pgesek:
The underlying reason is that @MotechListeners cannot be annotated
@Transactional at the same time. That leads to the listener being
registered two times - as the regular bean and as the transactional
proxy from Spring. This would cause issues with Commcare especially,
where the schema would get downloaded 2 times.

I improved the code a bit (create in Commcare wasn't doing any cleanup
for example). The doIntransaction from schedule-tracking can get improved
with better design, but didn't have time to look into that now.


pgesek:
Sebastian approved this pr for the master branch already

PR #208, opened 2016-03-14
MOTECH-2295 Fixed Case XML parsing for namespaces

sebbrudzinski:
The parsing of a case XML would not work properly if the XML sent to
MOTECH would contain elements with namespaces (eg. `
<n0:case> ...
</n0:case>
`, instead of `<case> ... </case>`). This has been fixed and the
parsing will work properly for elements with and without namespace.


sebbrudzinski:
Will also submit a PR against 0.28 and master, once it gets approved.

PR #210, opened 2016-03-15
MOTECH-2295: Include case type in minimal strategy events

pgesek:
No point in just having a generic trigger for the minimal strategy,
especially given that the regular task triggers wouldn't work. Don't
think we need the minimal strategy  this minimal


pgesek:
test this please

sebbrudzinski:
@pgesek I'm wondering if this is the best solution. Yes, the tasks will properly fire now with minimal strategy, but none of the fields will be available anyways, which is missleading. I'll see that my task executed, but none of the trigger fields were included in the actions. Wouldn't it be better to generate task triggers for case types only if the strategy is set full and only have one generic trigger for cases, if strategy is minimal or partial?

PR #213, opened 2016-03-15
MOTECH-2278 Console error removed

lkukielka:


commcare/src/main/resources/webapp/js/directives.js
```diff
@@ -431,7 +431,9 @@
 
                 if (scope.$parent.selectedConfig !== undefined) {
                     scope.downloadingCases = true;
-                    scope.$apply();
+                    if (!scope.$$phase) {
+                        scope.$apply();
+                    }
```
sebbrudzinski: According to Angular wiki, wrapping $apply into such condition is considered an anti-pattern. 
https://github.com/angular/angular.js/wiki/When-to-use-$scope.$apply%28%29

Perhaps we should investigate more throughly why $apply is called here and why we need it at all and do small refactor if necessary.

lkukielka:
According to http://www.codingeek.com/angularjs/angular-js-apply-timeout-digest-evalasync/ I used work around for this anti-pattern.

commcare/src/main/resources/webapp/js/directives.js
```diff
@@ -427,11 +427,10 @@
 
                 if (scope.$parent.selectedConfig) {
                     params.url = '../commcare/caseList/' + scope.$parent.selectedConfig.name + '?caseName=&dateModifiedStart=&dateModifiedEnd=';
-                }
-
-                if (scope.$parent.selectedConfig !== undefined) {
                     scope.downloadingCases = true;
-                    scope.$apply();
+                    setTimeout(function () {
```
pgesek: why not use the angular $timeout here? I believe that's mentioned in that link you provided

pgesek: Also, since we are on 1.2.X angular now, isn't the best way to this according to the article to use $evalAsync (I'm guessing call elem.jqgrid in $evalAsync)

lkukielka:
Why should we call elem.jqgrid in $evalAsync? I dont get it
Finally I used $timeout instead of setTimeout

pgesek:
Because I'm assuming it's the jqgrid call that requires the apply to be called before it executes? 

I'm now wondering how timing out the empty function is the correct fix here? Just how is this different from simply removing the $scope.apply part completely?

pgesek:
I guess timing out an empty function is fine in case that the jqgrid call does not require the apply call beforehand, since we only want to trigger apply safely. If jqgrid requires the apply called beforehand, then I guess this will randomly work or not, since the timeout call is asynchronous.

Does the jqgrid call require the apply?

lkukielka:
@pgesek That's right. I found out that sometimes calling apply before jqgrid may be necessary. So, do You think calling jqgrid inside $timeout, just as below, will be ok?

`              

```
          if (scope.$parent.selectedConfig) {
                params.url = '../commcare/caseList/' + scope.$parent.selectedConfig.name +  
                                  '?caseName=&dateModifiedStart=&dateModifiedEnd=';
                scope.downloadingCases = true;
            }

            $timeout(function() {
                elem.jqGrid(params);
            }, 0);
```

`

pgesek:
Yes, I suppose that should be fine. Is the function in $timeout is executed before or after the $apply?

lkukielka:
@pgesek Unfortunately It triggers apply at the end

pgesek:
if that's the case, we can use the promise it returns to sync the grid call I think. Something like $timeout(...).then(call grid here) ?

PR #219, opened 2016-03-21
MOTECH-2332: Change feed content value from Set to String

frankhuster:


frankhuster:
The build passed, the code review is all yours, @nickdotreid :grin:

PR #231, opened 2016-04-01
MOTECH-2315: Updated FormTriggerBuilder to ignore forms without XMLNS set

ngraczewski:


commcare/src/test/java/org/motechproject/commcare/util/DummyCommcareSchema.java
```diff
@@ -162,4 +162,39 @@
...
+        fields1.add(CASE_FIELD3);
+
+        FormSchemaQuestionJson questionJson7 = new FormSchemaQuestionJson();
+        questionJson7.setQuestionLabel("Last visit");
+        questionJson7.setQuestionValue(FORM_QUESTION4);
+
+        FormSchemaJson formSchemaJson6 = new FormSchemaJson();
+        formSchemaJson6.setFormNames(formNames6);
+        formSchemaJson6.setQuestions(Collections.singletonList(questionJson7));
+        formSchemaJson6.setXmlns(null);
```
pgesek: The whole point of this test is this one line - we should expose it more. Either add a comment, or preferably change the variable name for formSchemaWithNullXmlns

PR #244, opened 2016-04-12
MOTECH-2346: Fixed: It's not possible to save Commcare settings after change Configuration Name

pkornowski:
-  Commcare configurations on the server are identified by name. If we changed name of Configuration and click "Save" button, then on the server we searching configuration identified by new name, which it wasn't yet. I added new parameter "oldName" which stores name before change. When we update our configuration, then on the server side, configuration is searching by not updated name.
-  When we change "Event Strategy" of commcare settings we get error in console about $scope.saveSettings is not exists. It was because of "Event strategy" parameter had autosave, which invoked saveSettings function. It's not neccessary, because this parameter is saving like other parameter during clicks "Save" button, so I deleted this autosave.


kmadej:
Do not add field 'oldName' to Config class, just send this name as parameter

Additionally please add a unit test.

commcare/src/main/java/org/motechproject/commcare/config/Configs.java
```diff
@@ -132,8 +132,8 @@ public boolean hasDefault() {
      *
      * @param config  the update source
      */
-    public void updateConfig(Config config) {
-        deleteConfig(config.getName());
+    public void updateConfig(Config config, String oldName) {
```
kmadej, line 136: add javadoc refering new parameter

commcare/src/main/java/org/motechproject/commcare/web/ConfigController.java
```diff
@@ -53,8 +54,8 @@ public void makeDefault(@RequestBody String name) {
     @ResponseStatus(HttpStatus.OK)
     @RequestMapping(method = RequestMethod.POST)
     @ResponseBody
-    public Config saveConfig(@RequestBody Config config) throws CommcareConnectionFailureException {
-        return configService.saveConfig(config);
+    public Config saveConfig(@RequestBody Config config, @RequestParam("oldName") String oldName) throws CommcareConnectionFailureException {
```
kmadej: this parameter in RequestParam annotation is redundant

commcare/src/test/java/org/motechproject/commcare/it/CommcareTasksIntegrationBundleIT.java
```diff
@@ -313,8 +314,8 @@ private HttpResponse sendMockForm() throws IOException, InterruptedException {
         return getHttpClient().execute(httpPost);
     }
 
-    private HttpResponse createConfiguration(Config config) throws IOException, InterruptedException {
-        HttpPost httpPost = new HttpPost(String.format("http://localhost:%d/commcare/configs", PORT));
+    private HttpResponse createConfiguration(Config config, String oldName) throws IOException, InterruptedException {
+        HttpPost httpPost = new HttpPost(String.format("http://localhost:%d/commcare/configs?oldName=%s", PORT, oldName));
```
kmadej: why you used this request to create config ? I thought that there are two requests, one for create and second for update config.

commcare/src/test/java/org/motechproject/commcare/web/ConfigControllerTest.java
```diff
@@ -79,10 +79,15 @@ public void shouldSetDefault() {
     public void testSaveSettings() throws BundleException, CommcareAuthenticationException, CommcareConnectionFailureException {
 
         Config config = ConfigsUtils.prepareConfigThree();
+        String oldName = config.getName();
+        controller.saveConfig(config, oldName);
 
-        controller.saveConfig(config);
+        verify(configService).saveConfig(config, oldName);
+
+        config.setName("updatedName");
```
kmadej: apart from fact that this test is rather redundant, I do not understand your changes. Why after veryfing you are setting name in config and then again use verify ? It doesn't make sense

kmadej:
Additionally, please add an unit test to CommcareConfigServiceImplTest.

pkornowski:
I split saveConfig. If we adding new configuration, then on the server is sending only config now. If we updating configuration, then we sending config and old name of configuration. I added test for save updated config and for save new config in CommcareConfigServiceImplTest.

commcare/src/main/java/org/motechproject/commcare/service/CommcareConfigService.java
```diff
@@ -10,12 +10,21 @@
 public interface CommcareConfigService {
 
     /**
+     * Updates the given configuration.
+     *
+     * @param configuration  the configuration to be saved
+     * @param oldName   name of configuration before update
```
pgesek: the update

commcare/src/main/java/org/motechproject/commcare/service/CommcareConfigService.java
```diff
@@ -10,12 +10,21 @@
 public interface CommcareConfigService {
 
     /**
+     * Updates the given configuration.
```
pgesek: Used for changing configuration names

commcare/src/main/java/org/motechproject/commcare/service/CommcareConfigService.java
```diff
@@ -10,12 +10,21 @@
...
+     *
+     * @param configuration  the configuration to be saved
+     * @param oldName   name of configuration before update
+     * @return  the saved configuration
+     */
+    Config saveUpdatedConfig(Config configuration, String oldName) throws CommcareConnectionFailureException;
```
pgesek: rename this method to updateConfig

commcare/src/main/java/org/motechproject/commcare/service/CommcareConfigService.java
```diff
@@ -10,12 +10,21 @@
...
+
+    /**
      * Creates the given configuration.
      *
      * @param configuration  the configuration to be saved
      * @return  the saved configuration
      */
-    Config saveConfig(Config configuration) throws CommcareConnectionFailureException;
+    Config saveNewConfig(Config configuration) throws CommcareConnectionFailureException;
```
pgesek: rename this method to addConfig

commcare/src/main/java/org/motechproject/commcare/service/impl/CommcareConfigServiceImpl.java
```diff
@@ -78,13 +78,42 @@ public CommcareConfigServiceImpl(@Qualifier("commcareAPISettings") SettingsFacad
     }
 
     @Override
-    public Config saveConfig(Config config) throws CommcareConnectionFailureException {
+    public Config saveUpdatedConfig(Config config, String oldName) throws CommcareConnectionFailureException {
```
pgesek: there is too much duplication between the methods - I might be wrong, but isn't saveNewConfig(config) the same as saveUpdatedConfig(config, config.getName()) ?

commcare/src/main/resources/webapp/js/controllers.js
```diff
@@ -247,6 +249,7 @@
             copy.forwardSchema = config.forwardSchema;
             copy.forwardStubs = config.forwardStubs;
             copy.forwardCases = config.forwardCases;
+            $scope.oldName = copy.name;
```
pgesek: don't like the idea of storing this in scope, while everything else is returned by the method. Can't we return an object of {copy, oldName} instead?

commcare/src/main/resources/webapp/js/controllers.js
```diff
@@ -506,9 +510,36 @@
             );
         };
 
-        $scope.saveConfig = function(element) {
+        $scope.saveUpdatedConfig = function(element) {
```
pgesek: updateConfig

commcare/src/main/resources/webapp/js/controllers.js
```diff
@@ -529,6 +560,15 @@
                 });
         };
 
+        $scope.saveConfig = function(element) {
+            blockUI();
+            if($scope.oldName!=="") {
```
pgesek: just change to if ($scope.oldName)

commcare/src/test/java/org/motechproject/commcare/service/impl/CommcareConfigServiceImplTest.java
```diff
@@ -63,4 +71,35 @@ public void shouldReturnBlankUrlForNullOrBlankServerUrl() {
...
+
+    @Test
+    public void shouldSaveNewConfiguration() throws CommcareConnectionFailureException {
+
+        Config config = configService.create();
+        config.setName("newName");
+
+        Config savedConfig = configService.saveNewConfig(config);
+        assertEquals(config.getName(), savedConfig.getName());
+    }
```
pgesek, line 81: new line after this

commcare/src/test/java/org/motechproject/commcare/service/impl/CommcareConfigServiceImplTest.java
```diff
@@ -6,19 +6,27 @@
...
+import java.util.ArrayList;
+import java.util.List;
 
 import static org.junit.Assert.assertEquals;
 import static org.mockito.Mockito.when;
 
 @RunWith(MockitoJUnitRunner.class)
 public class CommcareConfigServiceImplTest {
 
+
```
pgesek: no need for these empty lines

PR #246, opened 2016-04-13
Fixed Modules build

kmadej:
Now the motechquartz database is dropped before ITs.


sebbrudzinski:
We can have the motechquartz database dropped before ITs, but we should still fix the lack of backwards compatibility due to class name change.

kmadej:
Agree, I will push this migration to motech repo, however I think we can merge this one anyway.

PR #253, opened 2016-04-19
MOTECH-2356: Added getRelationship by person UUID to the OpenMRS data…

ngraczewski:
… provider

### DIFF
--- a/openmrs-19/src/main/java/org/motechproject/openmrs19/service/OpenMRSRelationshipService.java
+++ b/openmrs-19/src/main/java/org/motechproject/openmrs19/service/OpenMRSRelationshipService.java
@@ -1,0 +1,21 @@
+package org.motechproject.openmrs19.service;
+
+import org.motechproject.openmrs19.domain.Relationship;
+
+/**
+ * Interface for handling relationships on the OpenMRS server.
+ */
+public interface OpenMRSRelationshipService {
+
+    /**
+     * Returns relationship of the type with the given {@code typeUuid} for the person with the given
+     * {@code personUuid}.
+     *
+     * @param configName  the name of the configuration to be used while performing this action
+     * @param typeUuid  the UUID of the type
+     * @param personUuid  the UUID of the person
+     * @return the matching relationship
+     */
+    Relationship getByTypeUuidAndPersonUuid(String configName, String typeUuid, String personUuid);
+
+}
--- a/openmrs-19/src/main/java/org/motechproject/openmrs19/service/OpenMRSRelationshipService.java
+++ b/openmrs-19/src/main/java/org/motechproject/openmrs19/service/OpenMRSRelationshipService.java
@@ -2,20 +2,22 @@ package org.motechproject.openmrs19.service;
 
 import org.motechproject.openmrs19.domain.Relationship;
 
+import java.util.List;
+
 /**
  * Interface for handling relationships on the OpenMRS server.
  */
 public interface OpenMRSRelationshipService {
 
     /**
-     * Returns relationship of the type with the given {@code typeUuid} for the person with the given
+     * Returns a list of relationships of the type with the given {@code typeUuid} for the person with the given
      * {@code personUuid}.
      *
      * @param configName  the name of the configuration to be used while performing this action
      * @param typeUuid  the UUID of the type
      * @param personUuid  the UUID of the person
-     * @return the matching relationship
+     * @return the list o matching relationships
      */
-    Relationship getByTypeUuidAndPersonUuid(String configName, String typeUuid, String personUuid);
+    List<Relationship> getByTypeUuidAndPersonUuid(String configName, String typeUuid, String personUuid);
 
 }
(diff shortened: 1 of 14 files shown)


openmrs-19/src/main/java/org/motechproject/openmrs19/service/OpenMRSRelationshipService.java
```diff
@@ -0,0 +1,21 @@
...
+    /**
+     * Returns relationship of the type with the given {@code typeUuid} for the person with the given
+     * {@code personUuid}.
+     *
+     * @param configName  the name of the configuration to be used while performing this action
+     * @param typeUuid  the UUID of the type
+     * @param personUuid  the UUID of the person
+     * @return the matching relationship
+     */
+    Relationship getByTypeUuidAndPersonUuid(String configName, String typeUuid, String personUuid);
```
pgesek: Shouldn't we also have a version with no config name, that uses the default config, just like in Commcare?

openmrs-19/src/main/java/org/motechproject/openmrs19/service/impl/OpenMRSRelationshipServiceImpl.java
```diff
@@ -0,0 +1,63 @@
...
+
+                if (relationship.getPersonB() != null) {
+                    personBUuid = relationship.getPersonB().getUuid();
+                }
+
+                if (relationship.getRelationshipType() != null) {
+                    relationshipTypeUuid = relationship.getRelationshipType().getUuid();
+                }
+
+                return StringUtils.equals(personBUuid, personUuid)
```
pgesek, line 48: why are we checking the person uuid here? won't the fetch handle that, given that we already pass the uuid as aparameter?

openmrs-19/src/main/java/org/motechproject/openmrs19/service/impl/OpenMRSRelationshipServiceImpl.java
```diff
@@ -0,0 +1,63 @@
...
+                        && StringUtils.equals(relationshipTypeUuid, typeUuid);
+            }
+        }));
+
+        if (relationships.size() > 1) {
```
pgesek: is it out of place to have two sons for example? Or am I getting this wrong? If this is done stricly for tasks reasons, then this check belongs in the data provider class

openmrs-19/src/test/java/org/motechproject/openmrs19/tasks/OpenMRSTaskDataProviderTest.java
```diff
@@ -264,4 +276,48 @@ public void shouldReturnProviderForLookupGetProviderByUuid() {
         assertEquals(provider, object);
         verify(providerService).getProviderByUuid(null, "5");
     }
+
+    @Test
+    public void shouldReturnRelationshipForLookupGetRelationshipByTypeUuidAndPersonBUuid() {
```
pgesek, line 283: No IT test? We can create a separate ticket for it at the very least

openmrs-19/src/main/java/org/motechproject/openmrs19/service/OpenMRSRelationshipService.java
ngraczewski: Methods without config name as parameters have been removed. With @kmadej changes we're losing backward compatibility, so those methods became redundant. I couldn't think of any other use case for them.

openmrs-19/src/main/java/org/motechproject/openmrs19/service/impl/OpenMRSRelationshipServiceImpl.java
ngraczewski, line 48: As mentioned in the [MOTECH-2356](https://applab.atlassian.net/browse/MOTECH-2356), at this moment some of the relationship filters doesn't work on the OpenMRS server. As a temporary workaround, we're filtering relationships by its "person", which includes both personA and personB fields. We're interested only in the personB, hence we are checking that person UUID.

openmrs-19/src/main/java/org/motechproject/openmrs19/service/impl/OpenMRSRelationshipServiceImpl.java
ngraczewski: It was done strictly for task reasons. The check has been moved to the data provider class.

openmrs-19/src/test/java/org/motechproject/openmrs19/tasks/OpenMRSTaskDataProviderTest.java
ngraczewski, line 283: Whole OpenMRSTaskDataProvider could use some integration tests. I've created ticket for it [MOTECH-2431](https://applab.atlassian.net/browse/MOTECH-2431).

PR #267, opened 2016-05-04
MOTECH-2360: Created UI for Connection management under the OpenMRS tab

weronika-redlarska:


sebbrudzinski:
Must be merged with latest master

openmrs-19/src/main/java/org/motechproject/openmrs19/service/impl/OpenMRSConfigServiceImpl.java
```diff
@@ -35,6 +35,24 @@ public void postConstruct() {
...
+        updateConfigs();
+    }
+
+    @Override
+    public void saveAllConfigs(Configs configs) {
+        List<Config> allConfigs = configs.getConfigs();
+        if (configs.getByName(configs.getDefaultConfigName()) != null) {
+            this.configs.setDefaultConfigName(configs.getDefaultConfigName());
+        }
+        for (int i=0; i<configs.getConfigs().size(); i++) {
```
pgesek: use a for-each

openmrs-19/src/main/java/org/motechproject/openmrs19/service/impl/OpenMRSConfigServiceImpl.java
```diff
@@ -35,6 +35,24 @@ public void postConstruct() {
...
+    public void deleteAllConfigs() {
+        configs.getConfigs().clear();
+        configs.setDefaultConfigName(null);
+        updateConfigs();
+    }
+
+    @Override
+    public void saveAllConfigs(Configs configs) {
+        List<Config> allConfigs = configs.getConfigs();
+        if (configs.getByName(configs.getDefaultConfigName()) != null) {
```
pgesek: any reason we are doing only if there is a previous default config?

openmrs-19/src/main/java/org/motechproject/openmrs19/web/ConfigController.java
```diff
@@ -0,0 +1,45 @@
...
+    @ResponseBody
+    public Configs getConfigs() {
+        return configService.getConfigs();
+    }
+
+    @ResponseStatus(HttpStatus.OK)
+    @RequestMapping(method = RequestMethod.POST)
+    @ResponseBody
+    public void updateConfigs(@RequestBody Configs configs) {
+        configService.deleteAllConfigs();
```
pgesek, line 35: should probably wrap this in a transaction at the configService layer

openmrs-19/src/main/resources/webapp/js/controllers.js
```diff
@@ -0,0 +1,156 @@
...
+                if ($scope.configs.configs[i].name === $scope.configs.defaultConfigName) {
+                    return $scope.configs.configs[i];
+                }
+            }
+            return undefined;
+        };
+
+         $scope.getConfigs = function () {
+             OpenMRSConfig.get(
+             function success(data) {
```
pgesek: weird identation, function should be on the previous line

openmrs-19/src/main/resources/webapp/js/controllers.js
```diff
@@ -0,0 +1,156 @@
...
+                 $scope.configs = data;
+                 $scope.configs.configs.patientIdentifierTypeNames=[];
+                 $scope.configOutdated = false;
+                 if ($scope.selectedConfig === undefined) {
```
pgesek: just check if ($scope.selectedConfig)

openmrs-19/src/main/resources/webapp/js/controllers.js
```diff
@@ -0,0 +1,156 @@
...
+                'openMrsUrl':'',
+                'username':'',
+                'password':'',
+                'motechPatientIdentifierTypeName':'',
+                'patientIdentifierTypeNames':[]
+            };
+        };
+
+        $scope.hasValue = function(prop) {
+            return $scope.selectedConfig[prop] !== null &&
```
pgesek: this is basically equivalent to if ($scope.selectedConfig[prop]) - we do not need the method

openmrs-19/src/main/resources/webapp/js/controllers.js
```diff
@@ -0,0 +1,156 @@
...
+                && $scope.hasValue('motechPatientIdentifierTypeName');
+        };
+
+        $scope.saveAllowed = function() {
+            return $scope.validateConfig()
+                && $scope.configOutdated;
+        };
+
+        $scope.isNewConfig = function() {
+            for (var i=0; i<$scope.configs.configs.length; i++) {
```
pgesek: is declaring var in this line valid? you declared it outside the for declaration previously

openmrs-19/src/main/resources/webapp/js/controllers.js
```diff
@@ -0,0 +1,156 @@
+(function () {
+    'use strict';
+
+    /* Controllers */
+
+    var controllers = angular.module('openmrs19.controllers',[]);
+
+    controllers.controller('OpenMRSSettingsCtrl', function($scope, OpenMRSConfig) {
+
+        $scope.configs = undefined;
```
pgesek: would be safer to declare this as an empty array I think

openmrs-19/src/main/resources/webapp/js/controllers.js
```diff
@@ -0,0 +1,156 @@
...
+                $scope.configs.defaultConfigName = $scope.selectedConfig.name;
+            }
+            blockUI();
+            OpenMRSConfig.save($scope.configs,
+                function success() {
+                    $scope.configOutdated = false;
+                    unblockUI();
+                },
+                function failure() {
+                    unblockUI();
```
pgesek: we should get some kind of notification in case of a failure

openmrs-19/src/main/resources/webapp/js/controllers.js
```diff
@@ -0,0 +1,156 @@
...
+            $scope.configs.configs.splice($scope.configs.configs.indexOf($scope.selectedConfig), 1);
+            blockUI();
+            OpenMRSConfig.save($scope.configs,
+                function success() {
+                    $scope.selectedConfig = undefined;
+                    $scope.getConfigs();
+                    unblockUI();
+                },
+                function failure() {
+                    unblockUI()
```
pgesek: same as above, the user needs to know that this failed

openmrs-19/src/main/resources/webapp/js/controllers.js
```diff
@@ -0,0 +1,156 @@
...
+            $scope.configs.defaultConfigName = $scope.selectedConfig.name;
+            blockUI();
+            OpenMRSConfig.save($scope.configs,
+                function success() {
+                    $scope.selectedConfig = undefined;
+                    $scope.getConfigs();
+                    unblockUI();
+                },
+                function failure() {
+                    unblockUI();
```
pgesek: the user needs to know that this failed

openmrs-19/src/main/resources/webapp/partials/settings.html
```diff
@@ -0,0 +1,72 @@
...
+            <label>{{msg('openMRS.selectConfig')}}: </label>
+            <select class="largeCombobox form-control" ng-change="selectChanged()" ng-model="selectedConfig" ng-options="config.name for config in configs.configs">
+            </select>
+        </div>
+
+        <div class="btn-group form-group">
+            <a ng-click="addConfig()" type="button" class="form-group btn btn-success">
+                <i class="fa fa-plus"></i>{{msg('openMRS.addConfig')}}
+            </a>
+            <a ng-if="selectedConfig" ng-disabled="!saveAllowed()" ng-click="saveAllowed() && saveConfig()" type="button" class="btn btn-primary">
```
pgesek, line 14: does ng-enabled exist? if yes we could avoid the negations

weronika-redlarska, line 14: ng-enabled does not exist

PR #270, opened 2016-05-05
MOTECH-2290 Add info to the top-level modules README

llavoie:


llavoie:
@craigappl another one for you, thanks!

PR #273, opened 2016-05-11
MOTECH-2442: Changing Groovy TemplateEngine to StreamingTemplateEngine

matkwiatkowski:
There was an error with to big templates in Groovy. Whole template is
acctually copied into Groovy method code, so there is problem with 64K bit
method size limit in Java.


kmadej:
You should not change groovy template, please find another solution (maybe try to use xmlslurper)

matkwiatkowski:
I have found in Groovy docs, that xml files larger than 64K can be only served in StreamingTemplateEngine
[Groove docs](http://docs.groovy-lang.org/docs/next/html/documentation/template-engines.html)

ihe-interop/src/main/java/org/motechproject/ihe/interop/handler/helper/IHEActionHelper.java
```diff
@@ -66,9 +66,9 @@ public void handleAction(Map<String, Object> parameters) throws IOException, Cla
         }
 
         Byte[] templateData = (Byte[]) iheTemplateDataService.getDetachedField(cdaTemplate, TEMPLATE_DATA_FIELD_NAME);
-        XmlTemplateEngine xmlTemplateEngine = new XmlTemplateEngine();
-        Template xmlTemplate = xmlTemplateEngine.createTemplate(new String(ArrayUtils.toPrimitive(templateData)));
-        Writable writable = xmlTemplate.make(parameters);
+        StreamingTemplateEngine streamingTemplateEngine = new StreamingTemplateEngine();
```
kmadej, line 73: is there a possibility to test this ?

pgesek:
@1stmateusz change the name of this PR

ihe-interop/src/main/java/org/motechproject/ihe/interop/handler/helper/IHEActionHelper.java
matkwiatkowski, line 73: I've added unit test below

ihe-interop/src/test/java/org/motechproject/ihe/interop/handler/helper/IHEActionHelperTest.java
```diff
@@ -61,4 +68,24 @@ public void shouldThrowExceptionWhenRecipientNotFound() throws ClassNotFoundExce
...
+        params.put(Constants.RECIPIENT_NAME_PARAM, "sampleRecipient4");
+
+        byte[] bytes = IOUtils.toByteArray(getClass().getResourceAsStream("/large_template_sample.txt"));
+        Byte[] byteObjects = new Byte[bytes.length];
+        int i = 0;
+        for (byte b : bytes) {
+            byteObjects[i++] = b;
+        }
+
+        when(iheTemplateDataService.findByName("sampleTemplate4")).thenReturn(cdaTemplate);
```
kmadej: use real xml template file, this test fails with previous template because its not xml, we should have a situation when running this test with old template throws java too large method error

kmadej:
@1stmateusz please push test in another PR, we will merge this PR without test

PR #276, opened 2016-05-12
MOTECH-2360: Fixes after QA in UI for Connection management under the OpenMRS tab

weronika-redlarska:
Verify button added.
Fixed problems with "make default" button, it is not possible to make default unsaved config.
Now it is not possible to create configuration with already existing name, also whitespaces are not allowed in configuration name.
Fixed issues with configChange event.


openmrs-19/src/main/resources/webapp/css/openmrs.css
```diff
@@ -0,0 +1,33 @@
+.openmrs .connection-status {
+    margin-top: 10px;
+    padding: 12px;
+    border-radius: 5px;
+}
+
+.openmrs .connection-error {
```
sebbrudzinski: Can we use bootstrap classes for those?

openmrs-19/src/main/java/org/motechproject/openmrs19/service/impl/OpenMRSConfigServiceImpl.java
```diff
@@ -113,6 +124,27 @@ public Config getDefaultConfig() {
         return configs.getByName(configs.getDefaultConfigName());
     }
 
+    @Override
+    @Transactional
```
sebbrudzinski: This method doesn't interract with the database at all, I don't think we need transactional here.

sebbrudzinski:
Moreover, conflicts must be resolved.

sebbrudzinski:
test this please

sebbrudzinski:
test this please

openmrs-19/src/main/java/org/motechproject/openmrs19/service/impl/OpenMRSConfigServiceImpl.java
```diff
@@ -53,16 +63,17 @@ public void postConstruct() {
     public void deleteAllConfigs() {
         configs.getConfigs().clear();
         configs.setDefaultConfigName(null);
-        updateConfigs();
     }
 
     @Override
     @Transactional
     public void saveAllConfigs(Configs configs) {
```
sebbrudzinski: I didn't notice this earlier, but we are shadowing the configs field with this method parameter. Can we rename the method parameter to avoid it?

PR #278, opened 2016-05-13
MOTECH-2420 Complete program enrollment and update its state

martokarski:
I created action to update program enrollment state. This action should also complete the program enrollment. There are two ways to do this. First is to change state to the terminal state (it automatically trigger program enrollment). Second is to set complete date. I created second action to fully covered requirements.


martokarski:
test this please

martokarski:
test this please

kmadej:
After conversation with Craig we established that we need only one task action. Please correct this PR

openmrs-19/src/main/java/org/motechproject/openmrs19/tasks/impl/OpenMRSActionProxyServiceImpl.java
```diff
@@ -146,6 +147,27 @@ public void createProgramEnrollment(String configName, String patientUuid, Strin
...
     }
 
+    @Override
+    public void changeStateOfProgramEnrollment(String configName, String programEnrollmentUuid, DateTime programCompletedDate,
+                                               String stateUuid, DateTime startDate) {
+        ProgramEnrollment programEnrollment = new ProgramEnrollment();
+        programEnrollment.setUuid(programEnrollmentUuid);
+        programEnrollment.setDateCompleted(Objects.nonNull(programCompletedDate) ? programCompletedDate.toDate() : null);
+
+        if (!Strings.isNullOrEmpty(stateUuid)) {
```
kmadej: we are usually using StringUtils from apache.commons, this class is already imported here so I think you should use this

kmadej:
test this please

PR #288, opened 2016-05-18
MOTECH-2364: Upgraded OpenMRS module to OpenMRS 1.12 (for Bahmni).

pkornowski:
OpenMRS v1.12 'get encounter' returns encounter with provider stored on the list, but OpenMRS v1.9 returns encounter with single provider object . I implemented method which rewrites provider from list to single provider object. Sending encounter in both versions works the same way.

### DIFF
--- a/openmrs-19/src/main/java/org/motechproject/openmrs19/domain/Encounter.java
+++ b/openmrs-19/src/main/java/org/motechproject/openmrs19/domain/Encounter.java
@@ -18,8 +18,9 @@ public class Encounter {
     private Date encounterDatetime;
     private Patient patient;
     private Person provider;
-    private List<Person> encounterProviders;
     private List<Observation> obs;
+    //This field is new in OpenMRS since version 1.10
+    private List<Person> encounterProviders;
 
     /**
      * Default constructor.
--- a/openmrs-19/src/main/java/org/motechproject/openmrs19/domain/Encounter.java
+++ b/openmrs-19/src/main/java/org/motechproject/openmrs19/domain/Encounter.java
@@ -18,6 +18,7 @@ public class Encounter {
     private Date encounterDatetime;
     private Patient patient;
     private Person provider;
+    private List<Person> encounterProviders;
     private List<Observation> obs;
 
     /**
@@ -95,6 +96,14 @@ public class Encounter {
 
     public void setProvider(Person provider) {
         this.provider = provider;
+    }
+
+    public List<Person> getEncounterProviders() {
+        return encounterProviders;
+    }
+
+    public void setEncounterProviders(List<Person> encounterProviders) {
+        this.encounterProviders = encounterProviders;
     }
 
     public List<Observation> getObs() {
--- a/openmrs-19/src/main/java/org/motechproject/openmrs19/resource/impl/EncounterResourceImpl.java
+++ b/openmrs-19/src/main/java/org/motechproject/openmrs19/resource/impl/EncounterResourceImpl.java
@@ -28,19 +28,30 @@ public class EncounterResourceImpl extends BaseResource implements EncounterReso
     public Encounter createEncounter(Config config, Encounter encounter) {
         String requestJson = buildGsonWithAdapters().toJson(encounter);
         String responseJson = postForJson(config, requestJson, "/encounter?v=full");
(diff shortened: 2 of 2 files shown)


pgesek:
@pkornowski can you create a set up for running ITs on v1.12 ? Esentially, what I would want to have is to make the OpenMRS url configurable through maven somehow. Currently it points to 'openmrs', where our OpenMRS 1.9 Docker is linked. I want to be able to point to openmrs-v1_12, where v1.12 will be linked.

pgesek:
Additionally we will want to rename openmrs19 to just openmrs, but that is a separate commit (and you should message the dev list about it).

openmrs-19/src/main/java/org/motechproject/openmrs19/domain/Encounter.java
```diff
@@ -18,6 +18,7 @@
     private Date encounterDatetime;
     private Patient patient;
     private Person provider;
+    private List<Person> encounterProviders;
```
kmadej: just name this field providers and please add comment here that this field is new in OpenMRS since version 1.10 (or 1.12 - check it).

openmrs-19/src/main/java/org/motechproject/openmrs19/resource/impl/EncounterResourceImpl.java
```diff
@@ -80,4 +91,11 @@ private Gson buildGsonWithAdapters() {
                 .registerTypeAdapter(Observation.class, new Observation.ObservationSerializer())
                 .create();
     }
+
+    private void getProviderFromEncounterProviderList(Encounter encounter) {
```
kmadej: please add javadoc here instead of adding comment

openmrs-19/src/main/java/org/motechproject/openmrs19/resource/impl/EncounterResourceImpl.java
```diff
@@ -80,4 +91,11 @@ private Gson buildGsonWithAdapters() {
...
+        //When OpenMRS version is 1.12, then providers are stored on the list.
+        if(encounter.getProvider()==null && !encounter.getEncounterProviders().isEmpty()) {
```
kmadej: whitespaces and use CollectionUtils.isNotEmpty

kmadej:
can you add a unit and integration test in another PR ? A unit test should include json with new field

openmrs-19/src/main/java/org/motechproject/openmrs19/domain/Encounter.java
pkornowski: OpenMRS 1.12 returns json with list named encounterProviders. This name must be 'encounterProviders'

kmadej: Ok, I understand now, so just add a comment

openmrs-19/src/main/java/org/motechproject/openmrs19/resource/impl/EncounterResourceImpl.java
```diff
@@ -80,4 +92,16 @@ private Gson buildGsonWithAdapters() {
...
                 .create();
     }
+
+    /**
+     * Rewrites provider from the list to single provider object in encounter. Since OpenMRS version 1.10
+     * providers in encounter are stored on the list.
+     *
+     * @param encounter
+     */
+    private void getProviderFromEncounterProviderList(Encounter encounter) {
```
kmadej: rename this method to setProviderFromEncounterProviderList. Usually get methods should return a value, here you are setting a provider value in encounter object.

openmrs-19/src/main/java/org/motechproject/openmrs19/resource/impl/EncounterResourceImpl.java
```diff
@@ -80,4 +92,16 @@ private Gson buildGsonWithAdapters() {
                 .registerTypeAdapter(Observation.class, new Observation.ObservationSerializer())
                 .create();
     }
+
+    /**
+     * Rewrites provider from the list to single provider object in encounter. Since OpenMRS version 1.10
```
kmadej: Sets provider from the list to a single provider object in encounter. ... as a list

PR #293, opened 2016-05-19
MOTECH-2364: Renamed openmrs19 module to openmrs.

pkornowski:


kmadej:
Looks good to me, I will merge this after sending an email to dev list.
Please change also name in motech repo in admin module

PR #294, opened 2016-05-20
MOTECH-2449: removing all fasterxml.jacson dependencies in modules

matkwiatkowski:


matkwiatkowski:
metrics-json dependency was used only once for creating bean healthCheckModule and later that bean wasn't used in metrics module at all. I assume, that only purpose af that bean was to initialize fasterxml.jacson context which we removed from motech and it is a leftover. metrics-json has fasterxml.jacson dependency, which causes our codehaus.jackson serializer to fail. It was also a problem in DHIS2 module solved in [MOTECH-2435](https://github.com/motech/modules/pull/281).

kmadej:
You should also handle problem with batch module in this PR

kmadej:
test this please

PR #297, opened 2016-05-20
MOTECH-2360: Fixed buttons behaviour in OpenMRS UI and added warnings

weronika-redlarska:
- Fixed buttons behaviour.
- Added prompt for Motech ID field.
- Now it is possible to write OpenMRS URL in both formats:
  - "http://where.is/openmrs"
  - "http://where.is/openmrs/"
- Added warnings when selected configuration is not saved.


openmrs-19/src/main/java/org/motechproject/openmrs19/config/Configs.java
```diff
@@ -37,7 +37,10 @@ public void add(Config config) {
         if (getByName(config.getName()) != null) {
             throw new ConfigurationAlreadyExistsException(config.getName());
         }
-
+        String openMrsUrl = config.getOpenMrsUrl();
```
sebbrudzinski: I wonder if this is the best place to handle that. What if I set URL with '/' via file configuration? Can we instead handle it at the retrieval of the URL from the settings facade?

openmrs-19/src/main/resources/webapp/js/controllers.js
```diff
@@ -14,10 +14,50 @@
         $scope.newPatientIdentifierType = {};
         $scope.failureMessage = undefined;
         $scope.successMessage = undefined;
+        $scope.rollback = false;
 
-        $scope.getDefault = function() {
+        $scope.$watch('selectedConfig', function(newValue, oldValue) {
```
sebbrudzinski: All of this seem really messy. Why do we need to keep track of "originalConfig", "configOutdated" and what even is the "rollback" variable?

openmrs/src/main/java/org/motechproject/openmrs/config/Configs.java
```diff
@@ -37,7 +37,10 @@ public void add(Config config) {
         if (getByName(config.getName()) != null) {
             throw new ConfigurationAlreadyExistsException(config.getName());
         }
-
+        String openMrsUrl = config.getOpenMrsUrl();
```
sebbrudzinski: It shouldn't be required anymore

openmrs/src/main/java/org/motechproject/openmrs/service/impl/OpenMRSConfigServiceImpl.java
```diff
@@ -159,4 +163,11 @@ private void updateConfigs() {
         settingsFacade.saveRawConfig(OPEN_MRS_CONFIGS_FILE_NAME, resource);
         eventRelay.sendEventMessage(new MotechEvent(EventSubjects.CONFIG_CHANGE_EVENT));
     }
+
+    private void validateOpenMrsUrl(Config config) {
```
sebbrudzinski: We do not really validate the URL, just adjust it a little

sebbrudzinski:
@pgesek OK to merge

PR #305, opened 2016-05-31
MOTECH-2420 Create Integration Tests for OpenMRS Program Enrollment

martokarski:


maven.properties
```diff
@@ -15,3 +15,6 @@ motech.sql.dbtype=mysql
 org.osgi.service.http.port=8181
 tomcat.port=9090
 tomcat.shutdown.port=9005
+
+openmrs.version=9
```
pgesek: this shouldn't be valid

pgesek: I guess this is just for Maven, but use 1.9 to be concise

kmadej:
@martokarski  please add here including different openmrs config depends on OpenMRS server version to OpenMRS ITs.

martokarski:
@pgesek I changed 1.9 to 1_9 because it is used in package name, so dots are not allowed.
@kmadej Config depends on openmrs.version property now.

openmrs/src/test/resources/config1_10/openmrs-configs.json
```diff
@@ -0,0 +1,19 @@
+{
+  "configs" :
+  [
+    {
+      "name": "Test OpenMRS Server",
+      "openMrsVersion":"1.10",
```
kmadej: 1.10+

openmrs/pom.xml
```diff
@@ -234,7 +234,7 @@
                                     <overwrite>true</overwrite>
                                     <resources>
                                         <resource>
-                                            <directory>src/test/resources</directory>
+                                            <directory>src/test/resources/config${openmrs.version}</directory>
```
kmadej, line 239: please add info about this new property to example above

openmrs/src/test/resources/config1_10/openmrs-configs.json
kmadej: and name this config package 1.12

openmrs/src/test/java/org/motechproject/openmrs/it/version1_10/MRSProgramEnrollmentIT.java
```diff
@@ -0,0 +1,205 @@
+package org.motechproject.openmrs.it.version1_10;
```
kmadej: rename this also to 1_12, because on the CI we have a docker with version 1.12

kmadej:
test this please

PR #308, opened 2016-06-07
MOTECH-1734: Added field with protocols, that can be merged with dhis2 server address. Basic address validation.

mkruszynski:


motech-gerrit:

motech-gerrit:

motech-gerrit:

sebbrudzinski:
test this please

dhis2/src/main/resources/webapp/js/controllers.js
```diff
@@ -13,6 +13,8 @@
                 .success(function (response) {
                     $scope.settings = response;
                     $scope.originalSettings = angular.copy($scope.settings);
+                    $scope.options = [{name: "http://", id: 1}, {name: "https://", id: 2}, {name: $scope.msg('dhis2.web.settings.serverURI.other'), id: 3}];
```
pgesek: What's other?

dhis2/src/main/resources/webapp/js/controllers.js
```diff
@@ -35,6 +37,14 @@
             };
 
             $scope.submit = function () {
+                if(!$scope.settings.serverURI.match("^(ht|f)tps?:\/\/.+")) {
```
pgesek: why the ht|f part? Doesn't  the dhis module only use HTTP(s)?

dhis2/src/main/resources/webapp/js/controllers.js
```diff
@@ -35,6 +37,14 @@
...
+                    if($scope.selectedProtocol.id == 3) {
```
pgesek: If we give them a dropdown for the protocol, we can assume that they won't enter it in the url. Eventually we can strip the protocol part from the input.  

On a side note, we don't have dropdowns for the protocol part anywhere in Motech. If we go with this, we should create a ticket for making the url input consistent.

sebbrudzinski:
test this please

dhis2/src/main/resources/webapp/js/controllers.js
```diff
@@ -35,6 +35,13 @@
             };
 
             $scope.submit = function () {
+                if(!$scope.settings.serverURI.match("^https?:\/\/.+")) {
+                    if($scope.settings.serverURI.match("^.+:\/\/.+")) {
```
pgesek: So what are we doing here? If I enter ftp:// for example, it replaces it with http? This is a feature we do not really need, all we need is to check if this is false, if false add the http:// protocol.

mkruszynski: I just wanted to replace other protocols with "http://". If it isn't need, I'll check only for http(s) occurence. If protocol will be absent I'll add it.

pgesek:
test this please

PR #311, opened 2016-06-09
MOTECH-2618 Fixed the console error while selecting default configuration

shanmukhm:


motech-gerrit:

motech-gerrit:

motech-gerrit:

motech-gerrit:

commcare/src/main/resources/webapp/js/controllers.js
```diff
@@ -297,7 +297,7 @@
                 } else {
                     $scope.rollback = false;
                 }
-            } else if (newValue !== undefined && newValue.name === "") {
+            } else if (newValue && newValue.name === "") {
```
pgesek: While at it, can we change this to (newValue && newValue.name) ? I hate these === "", they just cause issues

shanmukhm: @pgesek we are checking whether newValue.name is empty. So, shouldn't it be (newValue && !newValue.name)?

PR #313, opened 2016-06-13
MOTECH-2581: Hidden action and data source for OpenMRS 1.10+ in 1.9

pkornowski:
Now is not possible to create task with action or data source from OpenMRS 1.10+ when we use OpenMRS version 1.9.

### DIFF
--- a/openmrs/src/main/java/org/motechproject/openmrs/tasks/builder/ActionBuilder.java
+++ b/openmrs/src/main/java/org/motechproject/openmrs/tasks/builder/ActionBuilder.java
@@ -30,6 +30,7 @@ public class ActionBuilder {
     private static final String CREATE_PROGRAM_ENROLLMENT = "Create Program Enrollment";
     private static final String CHANGE_PROGRAM_ENROLLMENT_STATE = "Change Program Enrollment State";
     private static final String OPENMRS_ACTION_PROXY_SERVICE = "org.motechproject.openmrs.tasks.OpenMRSActionProxyService";
+    private static final String OPENMRS_V1_12 = "1.10+";
 
     private OpenMRSConfigService configService;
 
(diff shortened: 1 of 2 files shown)


openmrs/src/main/java/org/motechproject/openmrs/tasks/builder/ActionBuilder.java
```diff
@@ -30,6 +30,7 @@
     private static final String CREATE_PROGRAM_ENROLLMENT = "Create Program Enrollment";
     private static final String CHANGE_PROGRAM_ENROLLMENT_STATE = "Change Program Enrollment State";
     private static final String OPENMRS_ACTION_PROXY_SERVICE = "org.motechproject.openmrs.tasks.OpenMRSActionProxyService";
+    private static final String OPENMRS_V1_12 = "1.10+";
```
kmadej: this constant name is a bit misleading

openmrs/src/main/java/org/motechproject/openmrs/tasks/builder/ActionBuilder.java
```diff
@@ -49,8 +50,10 @@ public ActionBuilder(OpenMRSConfigService configService) {
             actions.add(buildCreateEncounterAction(configName));
             actions.add(buildCreatePatientAction(configName));
             actions.add(buildUpdatePatientAction(configName));
-            actions.add(buildCreateProgramEnrollmentAction(configName));
-            actions.add(buildChangeStateOfProgramEnrollmentAction(configName));
+            if(config.getOpenMrsVersion().equals(OPENMRS_V1_12)) {
```
kmadej: if you compare two strings, always start with raw one, it prevents NPE. Or you can use always StringUtils. 

Apart from that, I'm consider how to handle this, because I suppose that in the future maybe there will be another openMRS version like 1.11,  2.0 and so on. So maybe here the best solution is to compare with 1.9 and only if equals to 1.9 we do not add this actions. Moreover it will be more readable if you move this condition into private method. Lastly, please move this few lines at the end of adding actions code

openmrs/src/main/resources/velocity.templates/task-data-provider.vm
```diff
@@ -209,7 +209,8 @@
               "fieldKey": "endDate"
             }
           ]
-        },
+        }
+        #if( $config.getOpenMrsVersion() == "1.10+" ),
```
kmadej: the same as above, I think that it will be better to compare with 1.9

kmadej:
test this please

PR #315, opened 2016-06-15
MOTECH-2502: Audit Field Migration

mgrabowski7:


motech-gerrit:

motech-gerrit:

motech-gerrit:

http-agent/src/main/resources/db/migration/V1_MOTECH-2502.sql
```diff
@@ -0,0 +1,3 @@
+-- change type of field "responseBody" to text
```
sebbrudzinski: Shouldn't we have two migration files? One for mysql and one for postgresql?
Is this a correct path for the migration? Aren't they placed in db/migration/mysql and db/migration/default?

http-agent/src/main/resources/db/migration/V1_MOTECH-2502.sql
```diff
@@ -0,0 +1,3 @@
...
+
+ALTER TABLE "HTTP_AGENT_MODULE_HTTPACTIONAUDIT" ALTER COLUMN "responseBody" TYPE text;
```
sebbrudzinski: From what I recall this was supposed to be mediumtext or longtext in mysql. Can you check what type is created on clean install w/o the migration when we set the field type as text and use that db type in migration?

http-agent/src/main/java/org/motechproject/http/agent/domain/HTTPActionAudit.java
```diff
@@ -17,7 +17,7 @@
     @Field
     private String request;
 
-    @Field
+    @Field(type = "TEXT")
```
sebbrudzinski: We have a constant for that already.

http-agent/src/main/resources/db/migration/V1_MOTECH-2502.sql
sebbrudzinski: @mgrabowski7 Actually, this entity has been just recently added, so we can skip writing a specific migration to fix that column type.

sebbrudzinski:
test this please

http-agent/src/main/java/org/motechproject/http/agent/domain/HTTPActionAudit.java
```diff
@@ -17,7 +17,7 @@
     @Field
     private String request;
```
sebbrudzinski, line 18: Btw. can we also make the request a TEXT type? 255 characters is not that much and I can easily imagine someone complaining about it soon.

sebbrudzinski:
test this please

PR #326, opened 2016-07-07
MOTECH-2504: Fixed logging of HTTP-Agent module

nkowalik:


motech-gerrit:

motech-gerrit:

motech-gerrit:

motech-gerrit:

http-agent/src/main/java/org/motechproject/http/agent/listener/HttpClientEventListener.java
```diff
@@ -200,6 +200,13 @@ public void handle(MotechEvent motechEvent) {
         try {
             retValue = r.call();
         } catch (Exception e) { // Http request failed for all retries
+            if (e instanceof HttpException && !(e.getCause().toString().contains("I/O error"))) {
```
sebbrudzinski, line 203: We could just catch the HttpException, instead of generic Exception

sebbrudzinski, line 203: cause can be null - potential NPE

sebbrudzinski, line 203: Verifying whether a string representation of the cause exception contains something is wrong

http-agent/src/main/java/org/motechproject/http/agent/listener/HttpClientEventListener.java
```diff
@@ -200,6 +200,13 @@ public void handle(MotechEvent motechEvent) {
...
+                String cause = e.getCause().toString();
```
sebbrudzinski, line 204: Another potential NPE

http-agent/src/main/java/org/motechproject/http/agent/listener/HttpClientEventListener.java
```diff
@@ -200,6 +200,13 @@ public void handle(MotechEvent motechEvent) {
...
+                cause = cause.substring(cause.indexOf(":") + 2, cause.indexOf(":") + 5);
```
sebbrudzinski, line 205: Crafting the cause like this is super unintuitive. Moreover we cannot predict all exception causes, so this can produce anything.

sebbrudzinski, line 205: Is it even used anywhere?

sebbrudzinski, line 205: The whole logic here seems broken/invalid. 
Where does this exception even come from? 
Do we throw it ourselves? 
Is the response object available at some earlier point in the code so we could make the log?

PR #332, opened 2016-07-13
MOTECH-2649a: remove deleting task after test

matkwiatkowski:
deleting task after test sometimes causes tasks module failure

### DIFF
--- a/openmrs/src/test/java/org/motechproject/openmrs/it/version1_12/MRSTaskIntegrationBundleIT.java
+++ b/openmrs/src/test/java/org/motechproject/openmrs/it/version1_12/MRSTaskIntegrationBundleIT.java
@@ -98,8 +98,6 @@ public class MRSTaskIntegrationBundleIT extends AbstractTaskBundleIT {
     private static final Integer MAX_RETRIES_BEFORE_FAIL = 20;
     private static final Integer WAIT_TIME = 2000;
 
-    private Long taskID;
-
     @Override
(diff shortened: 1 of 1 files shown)


openmrs/src/test/java/org/motechproject/openmrs/it/version1_12/MRSTaskIntegrationBundleIT.java
```diff
@@ -289,9 +284,5 @@ private boolean checkIfProgramEnrollmentWasCreatedProperly() {
     private void activateTrigger() {
         settingsDataService.create(new SettingsRecord());
     }
-
-    private void deleteTask(Long taskID) {
```
pgesek: the current solution of stoing taskId as an instance variable is pointless, but why are we getting rid of task removal altogether?

matkwiatkowski: There is a problem with deleting task after test execution. When action was executed successfully task activity is stored in database and after checking it test ends and task is deleted, but actually task still wants to store some data in task record, despite test has executed properly. The one way of solving this is to add timeout before deleting the task, but I've realised, that deleting task is only needed while debugging test several times locally because database is cleared before test so I just deleted it.

kmadej:
test this please

kmadej:
test this please

PR #337, opened 2016-07-18
MOTECH-2228: fix retreving data in dhis2 release 2.22

matkwiatkowski:
- there was a change in retrieve list of objects in DHIS 2.22 release, now server replies only with id and display name instead of full data object, especially there was lack of href attribute so it was impossible to retrieve detailed information about object in list; to get full object information while getting list of them api requires to add "?fields=:identifiable" filter at the end of URL, another solution is to get information for every single object
- also fixed typo in DHIS2 contollers.js, that causes error in browser console and empty table of Tracked Entities

for checking api changes view [2.22](https://www.dhis2.org/222-upgrade)

### DIFF
--- a/dhis2/src/main/java/org/motechproject/dhis2/rest/domain/ProgramStageDataElementDto.java
+++ b/dhis2/src/main/java/org/motechproject/dhis2/rest/domain/ProgramStageDataElementDto.java
@@ -10,7 +10,7 @@ import java.util.Objects;
  */
 @JsonSerialize(include = JsonSerialize.Inclusion.NON_NULL)
 @JsonIgnoreProperties(ignoreUnknown = true)
-public class ProgramStageDataElementDto {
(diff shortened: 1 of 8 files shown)


pgesek:
@PiotrZurawskiSolDevelo please checks this with the different dhis viersions

PiotrZurawskiSolDevelo:
@pgesek @mkwiatkowskisoldevelo  i checked this.
I have possibility to Sync the DHIS2 MOTECH module with the DHIS2 application in version 2.20(older version), 2.22, 2.23 and 2.24.
Sync is working correctly but I can not see statement about successfully Synced application.

matkwiatkowski:
@PiotrZurawskiSolDevelo now sync message displays on UI. Please verify this.

PiotrZurawskiSolDevelo:
I checked this and now everything is working correctly.

PR #338, opened 2016-07-21
MOTECH-2738 Clear cookies before sending request

martokarski:
OpenMRS use basic HTTP authentication, but it also creates a cookie to keep user session. However this cookie doesn't have an expiration time. If server discards session, any request with session cookie will return "403" code. Our requests use basic authentication, so they don't need session cookie at all. To avoid problems with session timeout, I remove all cookies and rely only on basic authentication.

### DIFF
--- a/openmrs/src/main/java/org/motechproject/openmrs/resource/impl/BaseResource.java
+++ b/openmrs/src/main/java/org/motechproject/openmrs/resource/impl/BaseResource.java
@@ -89,6 +89,7 @@ public abstract class BaseResource {
     private ResponseEntity<String> exchange(Config config, URI url, HttpMethod method, String body, HttpHeaders headers) {
         headers.add("Authorization", "Basic " + prepareCredentials(config));
 
+        //Clear cookies to avoid SESSIONID mismatch
         httpClient.getState().clearCookies();
 
         return restOperations.exchange(url, method, new HttpEntity<>(body, headers), String.class);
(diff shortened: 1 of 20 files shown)


openmrs/src/main/java/org/motechproject/openmrs/resource/impl/BaseResource.java
```diff
@@ -85,6 +88,9 @@ protected void delete(Config config, String path, Object... params) {
 
     private ResponseEntity<String> exchange(Config config, URI url, HttpMethod method, String body, HttpHeaders headers) {
         headers.add("Authorization", "Basic " + prepareCredentials(config));
+
+        httpClient.getState().clearCookies();
```
kmadej, line 93: I think better idea will be remove cookie header from request or exclude using cookies by httpClient during initialization. Please do some research in this direction

martokarski, line 93: HttpClient class has deprecated method to set cookies policy, but it doesn't cover our requirements. I could use newer way to setup HttpClient, but this require another dependency.

kmadej, line 93: Ok thx for research. I think it will be better to avoid changing versions or adding new dependency just before 0.29 release. Especially that httpClient/RestTemplate is really often used in OpenMRS module. So far this solution looks good to me. However we should consider sth better in the future

kmadej:
test this please

kmadej:
test this please

PR #363, opened 2016-08-03
MOTECH-2800 Program Enrollment include attributes

martokarski:
Updated Program Enrollment and added new methods in Resource to support attributes in Program Enrollment. These methods will be used in:
- https://applab.atlassian.net/browse/MOTECH-2800
- https://applab.atlassian.net/browse/MOTECH-2798


kmadej:
Add also endpoint for creating Program Enrollment for : https://applab.atlassian.net/browse/MOTECH-2799

PR #396, opened 2016-08-19
Fixed constant key value

filgomul:
Keys constants should not contain dots in value. This will cause error i.e. while trying to get the value of trigger's bubble with such key.


motech-gerrit:

motech-gerrit:

motech-gerrit:

kmadej:
test this please

PR #408, opened 2016-08-26
MOTECH-2863: Fixed trigger subject name

mkruszynski:


openmrs/src/main/java/org/motechproject/openmrs/tasks/builder/OpenMRSTriggerBuilder.java
```diff
@@ -45,7 +45,8 @@ public OpenMRSTriggerBuilder(OpenMRSConfigService configService) {
 
         String displayName = DisplayNameHelper.buildDisplayName(DisplayNames.COHORT_QUERY_REPORTMEMBER, configName);
 
-        triggers.add(new TriggerEventRequest(displayName, EventSubjects.GET_COHORT_QUERY_MEMBER_EVENT, null, parameterRequests));
+        triggers.add(new TriggerEventRequest(displayName, EventSubjects.GET_COHORT_QUERY_MEMBER_EVENT.concat(".").concat(configName),
```
kmadej: you should also change event subject in 'Get CohortQuery Task Action'. Moreover, this provider and mentioned action should use one constant.

PR #413, opened 2016-08-31
MOTECH-2857 Dhis2 module connection leak - work I have done until now

filgomul:
This PR is a proposition to fix bug mentioned in Motech Developer Google Group (https://groups.google.com/forum/#!topic/motech-dev/egp1KWfvyM0) but it is not finished!

As Paweł and Samantha said in discussion, I tried to reproduce the error with mocked server acting as dhis2 one that just goes into a endless loop without closing the connection.

To reproduce the exception I did as followed:
- created task with Create Tracked Entity Instance action
- multiplied (through changes in code) number of created events when executing new task trigger (for example, I put https://github.com/filgomul/motech/blob/274cb4274d3d6d7eb4e10c1fa27c936afcc244d5/platform/mds/mds/src/main/java/org/motechproject/mds/service/DefaultMotechDataService.java#L144 in a loop and set CREATE MotechPermission as a trigger).

In createTrackedEntityInstance function there is commented out code showing how redirect request to the fake server.

I managed to get blocking request problem, and setting socket timeout seems to deal with this problem rather okay.

I couldn't fix the problem with CLOSE_WAIT connections Samantha mentioned in first post. There was no problem while working with dhis2 server, but consuming responses and closing requests didn't seem to work on SimpleHttpServer.

### DIFF
--- a/dhis2/src/main/java/org/motechproject/dhis2/rest/service/impl/DhisWebServiceImpl.java
+++ b/dhis2/src/main/java/org/motechproject/dhis2/rest/service/impl/DhisWebServiceImpl.java
@@ -2,20 +2,26 @@ package org.motechproject.dhis2.rest.service.impl;
 
 import com.fasterxml.jackson.databind.JavaType;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.sun.net.httpserver.HttpExchange;
+import com.sun.net.httpserver.HttpHandler;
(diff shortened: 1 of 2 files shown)


motech-gerrit:

motech-gerrit:

motech-gerrit:

motech-gerrit:

filgomul:
I did not mean to create PR on this repostiry so I am closing it.

PR #415, opened 2016-09-06
MOTECH-2857 Fix Dhis2 module connection leak

palbecki:
- deal with blocking request problem by setting socket timeout 
- deal with CLOSE_WAIT by consuming and close responses
- add deamon for closing expired connections

### DIFF
--- a/dhis2/src/main/java/org/motechproject/dhis2/rest/service/impl/DhisWebServiceImpl.java
+++ b/dhis2/src/main/java/org/motechproject/dhis2/rest/service/impl/DhisWebServiceImpl.java
@@ -1,22 +1,28 @@
 package org.motechproject.dhis2.rest.service.impl;
 
 import org.apache.commons.lang.StringUtils;
+import com.sun.net.httpserver.HttpExchange;
+import com.sun.net.httpserver.HttpHandler;
+import com.sun.net.httpserver.HttpServer;
 import org.apache.http.Header;
-import org.apache.http.HttpResponse;
 import org.apache.http.HttpStatus;
 import org.apache.http.StatusLine;
 import org.apache.http.auth.AuthenticationException;
 import org.apache.http.auth.UsernamePasswordCredentials;
-import org.apache.http.client.HttpClient;
+import org.apache.http.client.methods.CloseableHttpResponse;
 import org.apache.http.client.methods.HttpGet;
 import org.apache.http.client.methods.HttpPost;
 import org.apache.http.client.methods.HttpUriRequest;
 import org.apache.http.client.protocol.HttpClientContext;
+import org.apache.http.config.SocketConfig;
 import org.apache.http.entity.StringEntity;
 import org.apache.http.impl.auth.BasicScheme;
+import org.apache.http.impl.client.CloseableHttpClient;
+import org.apache.http.impl.conn.PoolingHttpClientConnectionManager;
 import org.apache.http.osgi.services.HttpClientBuilderFactory;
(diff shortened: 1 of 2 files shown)


motech-gerrit:

motech-gerrit:

motech-gerrit:

kmadej:
test this please

dhis2/src/main/java/org/motechproject/dhis2/rest/service/impl/DhisWebServiceImpl.java
```diff
@@ -511,6 +548,57 @@ private DhisDataValueStatusResponse importDataValues(Settings settings, String u
...
+        }
+    }
+
+    @PreDestroy
+    public void cleanUp() throws IOException {
+        client.close();
+        poolingHttpClientConnectionManager.close();
+    }
+
+    private final class IdleConnectionMonitorThread extends Thread {
```
pgesek: we should rather implement runnable than extend Thread

dhis2/src/main/java/org/motechproject/dhis2/rest/service/impl/DhisWebServiceImpl.java
```diff
@@ -511,6 +548,57 @@ private DhisDataValueStatusResponse importDataValues(Settings settings, String u
...
+        private final HttpClientConnectionManager connMgr;
+        private volatile boolean shutdown;
+
+        private IdleConnectionMonitorThread(PoolingHttpClientConnectionManager connMgr) {
```
pgesek: To be honest all of this looks fairly complicated. What are we exactly doing here, making sure we disconnect idle connections? This looks suspiciously complex for something fairly common in my opinion.

dhis2/src/main/java/org/motechproject/dhis2/rest/service/impl/DhisWebServiceImpl.java
```diff
@@ -511,6 +548,57 @@ private DhisDataValueStatusResponse importDataValues(Settings settings, String u
...
+    }
+
+    private void closeResponse(CloseableHttpResponse response) {
+        try {
+            if (response != null) {
+                EntityUtils.consume(response.getEntity());
+                response.close();
+            }
+        } catch (IOException e) {
+            LOGGER.error("Error trying to close response: " + e.getMessage());
```
pgesek: we want to print to ex with  the stracktrace, so do LOGGER.error("Error trying to close response", e) instead.

dhis2/src/main/java/org/motechproject/dhis2/rest/service/impl/DhisWebServiceImpl.java
palbecki: I was following instructions from apache httpclient tutorial that say:
 "The stale connection check is not 100% reliable. The only feasible solution that does not involve a one thread per socket model for idle connections is a dedicated monitor thread used to evict connections that are considered expired due to a long period of inactivity"
https://hc.apache.org/httpcomponents-client-ga/tutorial/html/connmgmt.html 
2.5. Connection eviction policy

pgesek: Fine, this can stay. This highlights the issues people have with http components though - complicated to use.

kmadej:
test this please

kmadej:
test this please

PR #424, opened 2016-09-14
MOTECH-2861: Fixed modules after switching off datanucleus property detachAllOnCommit

afijal:
- removed unnecessary addAppointmentDuplicate test from AppointmentServiceBundleIT (appointment's exteralId is unique so this test is unnecessary, moreover, test has been passing only because instead of duplicating appointment- it has been updating it)
- fixed imports after moving enums to separate package
- partially fixed MTrainingServiceBundleIT (testCourseUpdateAddChapter fails due to updateCourseStructure from class CourseStructureServiceImpl not working correctly)
- temporarily turned off few ITs from mTraining module due to problem mentioned above


kmadej (changes requested):
test

kmadej (approved):

kmadej (changes requested):
test

kmadej:
@afijal please change PR topic, you have to fix module ITs because of switching off detachAllOnCommit property

kmadej (approved):

kmadej:
There are problems with ITs in mTraining module, we will fix it as soon as possible in another PR. Additionally one test in mTraining does not pass on psql build, so I will temporary switch it off

PR #428, opened 2016-09-15
MOTECH-2922: Travis post-push trigger

mkruszynski:


sebbrudzinski (changes requested):
Let's run modules ITs before deploy

PR #429, opened 2016-09-15
MOTECH-2717 form trigger subject is now created from combination of xmlns and app_id

PJSosnowski:
Previously trigger subject was created from xmlns, now its combination of xmlns and app_id


motech-gerrit:

motech-gerrit:

pgesek (changes requested):
testing new reviews

pgesek (approved):

pgesek (changes requested):
Tests need to be updated

pgesek (changes requested):
Plus shouldn't we do this for datasources and actions as well?

kmadej:
@PJSosnowski please change PR topic, I suppose that now you can save a task

PJSosnowski:
@pgesek I've only applied this combination to actions, because as far as I can see data source doesn't use xmlns

kmadej:
there is a new PR for this ticket https://github.com/motech/modules/pull/459

PR #431, opened 2016-09-16
MOTECH-2926: Cherry-pick Don't Send Observation If Observation Value is Null or Empty String

palbecki:


motech-gerrit:

motech-gerrit:

PR #432, opened 2016-09-16
MOTECH-2916 Removed trailing slash from server URI

mjazdzdewski:


motech-gerrit:

motech-gerrit:

dhis2/src/main/java/org/motechproject/dhis2/service/Settings.java
```diff
@@ -14,7 +14,11 @@
     public Settings() { }
 
     public Settings(String serverURI, String username, String password) {
-        this.serverURI = serverURI;
+        if (serverURI.endsWith("/")) {
```
ngraczewski: Use splice(0, -1) and inline if for more readability.

sebbrudzinski: Including this logic at the construct level is not the best idea. If I use setter instead of the constructor to provide the server URL this won't be invoked and thus manifest itself again.

sebbrudzinski (changes requested):
We also need a test that verifies this to avoid regression.

dhis2/src/main/java/org/motechproject/dhis2/service/impl/SettingsServiceImpl.java
```diff
@@ -60,6 +60,10 @@ private synchronized void loadSettings() {
             LOGGER.debug("Loading DHIS2 settings...");
             Gson gson = new Gson();
             settings = gson.fromJson(jsonText, Settings.class);
+            String URI = settings.getServerURI();
```
sebbrudzinski: local variable name should not be capitalized

dhis2/src/main/java/org/motechproject/dhis2/service/impl/SettingsServiceImpl.java
```diff
@@ -60,6 +60,10 @@ private synchronized void loadSettings() {
...
+            if(settings.getServerURI().endsWith("/")) {
```
sebbrudzinski: why using settings.getServerURI() if you aleeady have that in a variable? (1 line above this)

sebbrudzinski (changes requested):

dhis2/src/main/java/org/motechproject/dhis2/service/impl/SettingsServiceImpl.java
```diff
@@ -60,6 +60,10 @@ private synchronized void loadSettings() {
             LOGGER.debug("Loading DHIS2 settings...");
             Gson gson = new Gson();
             settings = gson.fromJson(jsonText, Settings.class);
+            String uri = settings.getServerURI();
```
kmadej: please use URL builder as requested in ticket description

dhis2/src/test/java/org/motechproject/dhis2/service/impl/SettingsServiceImplTest.java
```diff
@@ -0,0 +1,59 @@
...
+            when(settingsFacade.getRawConfig(SETTINGS_FILE_NAME)).thenReturn(is);
+            settingsService = new SettingsServiceImpl(settingsFacade, eventRelay);
+        } catch (IOException e) {
+            String message = "There was an error loading json from the DHIS2 settings.";
+            throw new JsonIOException(message, e);
+        }
+    }
+
+    @Test
+    public void shouldParseURI() {
```
kmadej: change test to 'shouldRemoveTrailingSlashes'

dhis2/src/main/java/org/motechproject/dhis2/service/impl/SettingsServiceImpl.java
kmadej: It should help in handling double slashes

kmadej (changes requested):

dhis2/src/main/java/org/motechproject/dhis2/service/impl/SettingsServiceImpl.java
```diff
@@ -60,6 +61,12 @@ private synchronized void loadSettings() {
             LOGGER.debug("Loading DHIS2 settings...");
             Gson gson = new Gson();
             settings = gson.fromJson(jsonText, Settings.class);
+            URI uri = new URI(settings.getServerURI());
```
mjazdzdewski: @kmadej I've added URI normalization for removal of multiple slashes, but it leaves trailing slash, so I still remove it manually.

dhis2/src/main/java/org/motechproject/dhis2/service/impl/SettingsServiceImpl.java
```diff
@@ -60,6 +61,14 @@ private synchronized void loadSettings() {
             LOGGER.debug("Loading DHIS2 settings...");
             Gson gson = new Gson();
             settings = gson.fromJson(jsonText, Settings.class);
+            if (settings.getServerURI() != null) {
+                URI uri = new URI(settings.getServerURI());
+                uri = uri.normalize();
+                String uriString = uri.toString();
+                if (uriString.endsWith("/")) {
```
kmadej: please move this to separate method 'removeTrailingSlash'. The code will be more readable

kmadej (changes requested):

dhis2/src/main/java/org/motechproject/dhis2/service/impl/SettingsServiceImpl.java
```diff
@@ -60,10 +62,28 @@ private synchronized void loadSettings() {
...
+
+    private void removeTrailingSlash() {
+        try {
+            URI uri = new URI(settings.getServerURI());
+            uri = uri.normalize();
+            String uriString = uri.toString();
+            if (uriString.endsWith("/")) {
+                settings.setServerURI(uriString.substring(0, uriString.length() - 1));
+            }
+        } catch (URISyntaxException e) {
```
kmadej: duplicated code, this method should throw this exception and method above should catch this

kmadej (changes requested):

dhis2/src/main/java/org/motechproject/dhis2/service/impl/SettingsServiceImpl.java
```diff
@@ -60,10 +62,28 @@ private synchronized void loadSettings() {
...
+                removeTrailingSlash();
+            }
         } catch (Exception e) {
             String message = "There was an error loading json from the DHIS2 settings.";
             LOGGER.debug(message);
             throw new JsonIOException(message, e);
         }
     }
+
+    private void removeTrailingSlash() {
```
kmadej: additionally method should return parsedURI and set logic should be done above. It is more readable solution

kmadej (changes requested):

kmadej (approved):

PR #434, opened 2016-09-19
MOTECH-2939: Allowed Sending Multiple Observations of the Same Type

MichalKus:
Now it is possible to add multiple values to the same type of Observation. Each value should be separate by comma.

### DIFF
--- a/openmrs/src/main/java/org/motechproject/openmrs/tasks/impl/OpenMRSActionProxyServiceImpl.java
+++ b/openmrs/src/main/java/org/motechproject/openmrs/tasks/impl/OpenMRSActionProxyServiceImpl.java
@@ -264,8 +264,8 @@ public class OpenMRSActionProxyServiceImpl implements OpenMRSActionProxyService 
 
         for (String observationConceptName : observations.keySet()) {
             if (valueIsNotEmpty(observations, observationConceptName)) {
-                String[] observationsValues = observations.get(observationConceptName).replace(" ", "").split(",");
-                for (String value : observationsValues) {
+                String[] observationValues = observations.get(observationConceptName).replace(", ", ",").split(",");
+                for (String value : observationValues) {
                     Observation observation = new Observation();
 
                     ConceptName conceptName = new ConceptName(observationConceptName);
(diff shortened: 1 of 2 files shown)


motech-gerrit:

motech-gerrit:

motech-gerrit:

openmrs/src/main/java/org/motechproject/openmrs/tasks/impl/OpenMRSActionProxyServiceImpl.java
```diff
@@ -264,19 +264,21 @@ private Location getLocationByName(String configName, String locationName) {
 
         for (String observationConceptName : observations.keySet()) {
             if (valueIsNotEmpty(observations, observationConceptName)) {
-                Observation observation = new Observation();
+                String[] observationsValues = observations.get(observationConceptName).replace(" ", "").split(",");
```
kmadej: what if value string has space, we should not remove every space in string

kmadej: change variable name to observationValues

kmadej:
Please adjust/add unit test

kmadej (changes requested):
change

openmrs/src/test/java/org/motechproject/openmrs/tasks/impl/OpenMRSActionProxyServiceTest.java
```diff
@@ -174,6 +174,43 @@ public void shouldCreateEncounterWithGivenParametersWithoutObsWithEmptyValue() {
...
+
+        Person person = new Person();
+        person.setUuid("30");
+        provider.setPerson(person);
+
+        DateTime encounterDatetime = new DateTime("2000-08-16T07:22:05Z");
+        Map<String, String> observations = new HashMap<>();
+        observations.put("testConceptName","testObservationValueName0, testObservationValueName1");
+
+        List<Observation> obsList = createObservationListWithManyObservations();
```
kmadej: I think it would be  better to use existing method createObservationList and add a parameter, how many observations should be added

kmadej (changes requested):

kmadej (approved):

PR #437, opened 2016-09-21
MOTECH-2939: Allowed Sending Multiple Observations of the Same Type

MichalKus:
Now it is possible to add multiple values to the same type of Observation. Each value should be separate by comma.


motech-gerrit:

motech-gerrit:

PR #440, opened 2016-09-21
Motech 2885: Get most recent observation by patient and concept ID

mkruszynski:


pkornowski:
@kmadej Looks good to me.

kmadej:
On JIRA in ticket description there is the following information 'This data source can return 0 or 1 result. The data source bubbles are at the bottom of the ticket. Note, we should also be able to filter if 0 results are returned using a task filter'. So please handle this properly

kmadej:
Please resolve conflicts, and if data source returns 0, there should be possibility to filter this value. Now you return null, so it is not possible

kmadej:
The feature I mentioned above has still not been implemented

openmrs/src/main/java/org/motechproject/openmrs/service/impl/OpenMRSObservationServiceImpl.java
```diff
@@ -97,6 +98,20 @@ public Observation getObservationByUuid(String configName, String uuid) {
...
     @Override
+    public ObservationListResult getObservationByPatientUUIDAndConceptUUID(String configName, String patientUUID, String conceptUUID) {
+        Validate.notEmpty(patientUUID, "Patient uuid cannot be empty");
+        Validate.notEmpty(conceptUUID, "Concept uuid cannot be empty");
+
+        try {
+            Config config = configService.getConfigByName(configName);
+            return obsResource.getObservationByPatientUUIDAndConceptUUID(config, patientUUID, conceptUUID);
+        } catch (HttpClientErrorException e) {
+            LOGGER.error("Error while fetching observations with Patient uuid: " + patientUUID + " and Concept uuid: " + conceptUUID);
```
kmadej: throw exception with detailed error message

openmrs/src/main/java/org/motechproject/openmrs/tasks/OpenMRSTaskDataProvider.java
```diff
@@ -188,6 +198,29 @@ private Patient getPatient(String lookupName, Map<String, String> lookupFields,
...
+        ObservationListResult observations;
+        Observation latestObservation = null;
+
+        switch(lookupName) {
+            case BY_PATIENT_UUID_AND_CONCEPT_UUID:
+                observations = observationService.getObservationByPatientUUIDAndConceptUUID(configName,
+                        lookupFields.get(PATIENT_UUID), lookupFields.get(CONCEPT_UUID));
+                if (CollectionUtils.isNotEmpty(observations.getResults())) {
+                    latestObservation = observationService.getObservationByUuid(configName, observations.getResults().get(0).getUuid());
+                    latestObservation.setNumberOfObservations("1");
```
kmadej: please move this few lines into new method

openmrs/src/test/java/org/motechproject/openmrs/tasks/OpenMRSTaskDataProviderTest.java
```diff
@@ -107,7 +111,7 @@
     @Before
     public void setUp() {
         when(configService.getConfigs()).thenReturn(new Configs());
```
kmadej, line 90: please add unit test

openmrs/src/test/java/org/motechproject/openmrs/resource/impl/ObservationResourceImplTest.java
```diff
@@ -78,6 +79,23 @@ public void shouldQueryForObservationByPatientId() throws Exception {
...
+        when(restOperations.exchange(eq(url), eq(HttpMethod.GET), any(HttpEntity.class), eq(String.class)))
+                .thenReturn(getResponseFromFile(OBSERVATION_QUERY_RESPONSE_JSON));
+
+        ObservationListResult result = observationResource.getObservationByPatientUUIDAndConceptUUID(config, patientId, conceptId);
+
+        verify(restOperations).exchange(eq(url), eq(HttpMethod.GET), requestCaptor.capture(), eq(String.class));
+
+        assertThat(result, equalTo(readFromFile(OBSERVATION_QUERY_RESPONSE_JSON, ObservationListResult.class)));
+        assertThat(requestCaptor.getValue().getHeaders(), equalTo(getHeadersForGet(config)));
+        assertThat(requestCaptor.getValue().getBody(), nullValue());
```
kmadej: please remove this assert, I know that other tests have this line, however this check doesnt have much sense

openmrs/src/main/java/org/motechproject/openmrs/resource/impl/ObservationResourceImpl.java
```diff
@@ -44,8 +44,16 @@ public void voidObservation(Config config, String id, String reason) {
 
     @Override
     public Observation getObservationById(Config config, String uuid) {
+        Map<Type, Object> adapter = new HashMap<>();
```
kmadej, line 47: why do you change this method ?

openmrs/src/main/java/org/motechproject/openmrs/domain/Observation.java
```diff
@@ -211,19 +278,23 @@ public JsonElement serialize(ObservationValue src, Type typeOfSrc, JsonSerializa
...
+            if (json.isJsonObject()) {
+                JsonObject jsonObject = json.getAsJsonObject();
+                observationValue.setDisplay(jsonObject.get(DISPLAY_KEY).getAsString());
+            } else if (json.isJsonPrimitive()) {
+                JsonPrimitive jsonPrimitive = json.getAsJsonPrimitive();
+                observationValue.setDisplay(jsonPrimitive.getAsString());
+            }
 
-            return (ObservationValue) JsonUtils.readJson(valueObject.toString(), ObservationValue.class);
+            return observationValue;
```
kmadej, line 297: I see that you are using this deserializer only in method not connected with this data provider. What's the reason ?

kmadej (changes requested):

openmrs/src/main/java/org/motechproject/openmrs/domain/Observation.java
mkruszynski: When I was resolving conflicts after branch update I met conflicts in ObservationValue json Deserializer. I'll check that this change is necessary.

mkruszynski: @kmadej I checked it and this function is used to get latest observation. Deserializer of ObservationValue is needed, because this value in response json can be a primitive or json object.

kmadej (approved):

PR #441, opened 2016-09-21
MOTECH-2920: Used Concept UUID for Submitting Observations

MichalKus:
Now to add new Observation it is required to use Concept Uuid as a key.


motech-gerrit:

motech-gerrit:

PR #444, opened 2016-09-22
MOTECH-2939: Fixed blank/null values in Observation of the same type

MichalKus:
Now if value is blank/null Observation is not created.


motech-gerrit:

motech-gerrit:

PR #446, opened 2016-09-22
MOTECH-2904: hotfix change `visit type uuid` to `visit type`

palbecki:
- Visit Type in visit action can be `name` as well as `uuid` and i changed `visit type uuid` to `visit type`.


motech-gerrit:

motech-gerrit:

kmadej:
We should stay with uuid only, If needed we can handle this later. In the ticket description there is info about uuid.

PR #447, opened 2016-09-22
MOTECH-2915: Failing task when create encounter gets 400 error

nkowalik:


motech-gerrit:

motech-gerrit:

openmrs/src/main/java/org/motechproject/openmrs/service/impl/OpenMRSEncounterServiceImpl.java
```diff
@@ -70,8 +71,7 @@ public Encounter createEncounter(String configName, Encounter encounter) {
 
             eventRelay.sendEventMessage(new MotechEvent(EventKeys.CREATED_NEW_ENCOUNTER_SUBJECT, EventHelper.encounterParameters(createdEncounter)));
         } catch (HttpClientErrorException e) {
-            LOGGER.error("Could not create encounter: " + e.getMessage());
-            return null;
+            throw new OpenMRSException("Could not create encounter with patient uuid: " + encounter.getPatient().getUuid(), e);
```
kmadej: I am not sure if information we need will be displayed here. Can you verify this ? We should have here exact informaction what is wrong when 400 returned. I think getResponseBodyAsString should be used here.

kmadej (changes requested):

openmrs/src/main/java/org/motechproject/openmrs/service/impl/OpenMRSEncounterServiceImpl.java
```diff
@@ -58,8 +59,7 @@ public Encounter createEncounter(String configName, Encounter encounter) {
 
             eventRelay.sendEventMessage(new MotechEvent(EventKeys.CREATED_NEW_ENCOUNTER_SUBJECT, EventHelper.encounterParameters(createdEncounter)));
         } catch (HttpClientErrorException e) {
-            LOGGER.error("Could not create encounter: " + e.getMessage());
-            return null;
+            throw new OpenMRSException("Could not create encounter with patient uuid: " + encounter.getPatient().getUuid() + " " + e.getMessage() + " " + e.getResponseBodyAsString(), e);
```
kmadej: please use String.format. We should avoid concatting strings, because it creates and stores multiple strings in Java String pool

kmadej (changes requested):

kmadej (approved):

PR #449, opened 2016-09-22
MOTECH-2918 Added DHIS2 status response to DEBUG log and throwing DhisWebException when response is ERROR.

mjazdzdewski:


motech-gerrit:

motech-gerrit:

dhis2/src/main/java/org/motechproject/dhis2/rest/service/impl/DhisWebServiceImpl.java
```diff
@@ -420,6 +421,14 @@ private DhisStatusResponse createEntity(Settings settings, String uri, String js
             closeResponse(response);
         }
 
+        LOGGER.debug(String.format("DHIS2 status response: %s", status.getStatus().toString()));
+
+        if (status.getStatus() == DhisStatus.ERROR) {
```
kmadej: Did you check it correctly ? In the ticket description on JIRA there is an information that response can return 200 (OK) but response body may contain 'ERROR', so it seems that you do not include this condition

kmadej (changes requested):

dhis2/src/main/java/org/motechproject/dhis2/rest/service/impl/DhisWebServiceImpl.java
```diff
@@ -420,6 +421,14 @@ private DhisStatusResponse createEntity(Settings settings, String uri, String js
...
+            String msg = String.format("Error in DHIS2 status response: %s", status.getStatus().toString());
```
kmadej: we should log here more information if possible. Ideally the reason why there is a error. Additionally this 'status.getStatus().toString()' is redundant, because in this condition there will be always ERROR type

kmadej (changes requested):

dhis2/src/main/java/org/motechproject/dhis2/rest/service/impl/DhisWebServiceImpl.java
```diff
@@ -411,7 +413,14 @@ private DhisStatusResponse createEntity(Settings settings, String uri, String js
...
 
         try (InputStream content = getContentForResponse(response)) {
-            status = new ObjectMapper().readValue(content, DhisStatusResponse.class);
+            String contentString = IOUtils.toString(content);
+            status = new ObjectMapper().readValue(contentString, DhisStatusResponse.class);
+            if (status.getStatus() == DhisStatus.ERROR) {
+                LOGGER.debug(String.format("DHIS2 status response error details: %s", contentString));
+                String msg = String.format("Error in DHIS2 status response.");
+                statusMessageService.warn(msg, MODULE_NAME);
+                throw new DhisWebException(msg);
```
kmadej, line 420: 1. From JIRA 'Ideally task should fail properly with an error'. I see that task fails with exception and brief error message. We should extend this to fail properly with an error
2. Additionally as far as I understand, warn method from statusMessageService is logging this message. So now we are logging on debug level using LOGGER and then using service ? We should standardize this

kmadej (changes requested):

dhis2/src/main/java/org/motechproject/dhis2/rest/service/impl/DhisWebServiceImpl.java
```diff
@@ -411,7 +413,13 @@ private DhisStatusResponse createEntity(Settings settings, String uri, String js
         DhisStatusResponse status;
 
         try (InputStream content = getContentForResponse(response)) {
-            status = new ObjectMapper().readValue(content, DhisStatusResponse.class);
+            String contentString = IOUtils.toString(content);
+            status = new ObjectMapper().readValue(contentString, DhisStatusResponse.class);
+            if (status.getStatus() == DhisStatus.ERROR) {
+                String msg = String.format("DHIS2 status response error details: %s", contentString);
+                statusMessageService.warn(msg, MODULE_NAME);
```
kmadej: debug level

kmadej (changes requested):

dhis2/src/main/java/org/motechproject/dhis2/rest/service/impl/DhisWebServiceImpl.java
kmadej: I have read once again ticket description and I think that we should omit logging this, just throw exception and task will fail

kmadej (changes requested):

kmadej (approved):

PR #453, opened 2016-09-26
MOTECH-2917 Show Multiple Lines in Tasks Value Text Fields

wstrzelczyk:


wstrzelczyk:
test this please

kmadej:
I have pulled your changes and it seems that the problem still exists

PR #455, opened 2016-09-27
MOTECH-2949: Fixed random tests fails

mkruszynski:


kmadej:
It seems that there is still problem with Commcare IT and probably ODK IT as well

kmadej:
@mkruszynski   thx. I think we should restart travis job few times (let's do it 10 times) and check if this code fixes the issue

PR #460, opened 2016-10-04
MOTECH-2955 Event Strategy Should Defaulted to "Full"

srujan92:
Event Strategy default value has been set to 'Full'.


motech-gerrit:

motech-gerrit:

PR #461, opened 2016-10-05
MOTECH-2919: Improved OpenMRS logs for Http exceptions

mkruszynski:


kmadej:
Please verify OpenMRS ITs, I'm pretty sure that they will fail after this changes

kmadej:
You forgot about getConceptByUuid from OpenMRsConceptService. Please remember to verify ITs once again after changing exception handling

PR #463, opened 2016-10-13
MOTECH-2918 Added response details to debug log and error status handling to Data Values import

mjazdzdewski:
Added response details to debug log and error status handling to Data Values import

### DIFF
--- a/dhis2/src/main/java/org/motechproject/dhis2/rest/domain/DhisResponse.java
+++ b/dhis2/src/main/java/org/motechproject/dhis2/rest/domain/DhisResponse.java
@@ -1,0 +1,25 @@
+package org.motechproject.dhis2.rest.domain;
+
+/**
+ * Base class for DHIS2 action response.
+ */
+public abstract class DhisResponse {
+    private DhisStatus status;
+    private ImportCountDto importCountDto;
+
+    public DhisStatus getStatus() {
+        return status;
+    }
+
+    public void setStatus(DhisStatus status) {
+        this.status = status;
+    }
+
+    public ImportCountDto getImportCountDto() {
+        return importCountDto;
+    }
+
+    public void setImportCountDto(ImportCountDto importCountDto) {
+        this.importCountDto = importCountDto;
+    }
+}
--- a/dhis2/src/main/java/org/motechproject/dhis2/rest/domain/DhisResponse.java
+++ b/dhis2/src/main/java/org/motechproject/dhis2/rest/domain/DhisResponse.java
@@ -1,8 +1,11 @@
 package org.motechproject.dhis2.rest.domain;
 
+import org.codehaus.jackson.annotate.JsonIgnoreProperties;
+
 /**
  * Base class for DHIS2 action response.
  */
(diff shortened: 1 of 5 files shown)


motech-gerrit:

motech-gerrit:

dhis2/src/main/java/org/motechproject/dhis2/rest/service/impl/DhisWebServiceImpl.java
```diff
@@ -547,17 +546,24 @@ private DhisDataValueStatusResponse importDataValues(Settings settings, String u
...
 
         try (InputStream content = getContentForResponse(response)) {
-            return new ObjectMapper().readValue(content, DhisDataValueStatusResponse.class);
+            String contentString = IOUtils.toString(content);
+            LOGGER.debug(String.format("Received response to create resource: %s, request: %s", json, contentString));
+            status = new ObjectMapper().readValue(contentString, DhisDataValueStatusResponse.class);
+            if (status.getStatus() == DhisStatus.ERROR) {
+                String msg = String.format("Error in DHIS2 status response, error details: %s", contentString);
+                throw new DhisWebException(msg);
+            }
```
kmadej, line 545: can you create separate method for this, because this is almost the same code as above

kmadej (changes requested):

dhis2/src/main/java/org/motechproject/dhis2/rest/service/impl/DhisWebServiceImpl.java
```diff
@@ -547,17 +524,27 @@ private DhisDataValueStatusResponse importDataValues(Settings settings, String u
...
+    }
 
+    private Object getDhisResponseForHttpResponse(String json, String uri, CloseableHttpResponse response, Class responseType) {
+        Object status;
         try (InputStream content = getContentForResponse(response)) {
-            return new ObjectMapper().readValue(content, DhisDataValueStatusResponse.class);
+            String contentString = IOUtils.toString(content);
+            LOGGER.debug(String.format("Received response to create resource: %s, request: %s", contentString, json));
+            status = new ObjectMapper().readValue(contentString, responseType);
+            if (status.toString().contains("ERROR")) {
```
kmadej: you should use enum here instead of string. I think that the best solution will be creating new AbstractClass. DhisDataValueStatusResponse and DhisStatusResponse should extend this new class and then you can use here getStatus method

kmadej (changes requested):

dhis2/src/main/java/org/motechproject/dhis2/rest/domain/DhisResponse.java
```diff
@@ -0,0 +1,25 @@
+package org.motechproject.dhis2.rest.domain;
+
+/**
+ * Base class for DHIS2 action response.
+ */
+public abstract class DhisResponse {
```
kmadej, line 9: add JsonIgnoreProperties(ignoreUnknown = true)

kmadej (changes requested):

kmadej (approved):

PR #468, opened 2016-10-20
MOTECH-2959 Added create order action

kmadej:
- added create order action
- updated/added properties in openmrs
- added optional order field in create observation action


pkornowski:
@kmadej LGTM

PR #473, opened 2016-10-25
MOTECH-2959: Added unit tests for 'create order' action.

pkornowski:


mkruszynski:
@pkornowski Looks good to me

PR #481, opened 2016-11-21
MOTECH-2973 Import Forms By Form ID

srujan92:
1. Created a dropdown option, input field for "By Form Id" and added related js functions.
2. Added function in controller and created new methods for form import through id.
3. New test cases are added for the newly added methods.

### DIFF
--- a/commcare/src/main/java/org/motechproject/commcare/pull/CommcareFormImporter.java
+++ b/commcare/src/main/java/org/motechproject/commcare/pull/CommcareFormImporter.java
@@ -17,27 +17,52 @@ public interface CommcareFormImporter {
     /**
      * Counts how many forms match the provided criteria. A single request to the form list
      * API will be made underneath, and the returned total count will be returned by this method.
-     * @param dateRange the date range for the import
+     *
+     * @param dateRange  the date range for the import
      * @param configName the name of the configuration to use, if null is provided the default configuration
      *                   will be used
      * @return the total number of forms matching the provided criteria
      * @throws IllegalArgumentException if the date range is null or invalid (start date after end date)
-     * @throws IllegalStateException if there is an already ongoing import in progress
+     * @throws IllegalStateException    if there is an already ongoing import in progress
      */
     int countForImport(Range<DateTime> dateRange, String configName);
+
+    /**
+     * Checks if a form exists with select id. A single request to the form list
+     * API will be made underneath, and the returned total count will be returned by this method.
+     *
+     * @param formId     the id of the form to be checked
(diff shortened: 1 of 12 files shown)


motech-gerrit:

motech-gerrit:

commcare/src/main/java/org/motechproject/commcare/pull/CommcareFormImporter.java
```diff
@@ -17,35 +17,61 @@
...
      * @return the total number of forms matching the provided criteria
      * @throws IllegalArgumentException if the date range is null or invalid (start date after end date)
-     * @throws IllegalStateException if there is an already ongoing import in progress
+     * @throws IllegalStateException    if there is an already ongoing import in progress
      */
     int countForImport(Range<DateTime> dateRange, String configName);
 
     /**
+     * Checks if a form exists with select id. A single request to the form list
+     * API will be made underneath, and the returned total count will be returned by this method.
```
kmadej: this method does not return total count

commcare/src/main/java/org/motechproject/commcare/pull/CommcareFormImporter.java
```diff
@@ -17,35 +17,61 @@
...
+     *
+     * @param formId     the id of the form to be checked
+     * @param configName the name of the configuration to use, if null is provided the default configuration
+     *                   will be used
+     * @return the boolean value if there exists a form with the provided id
+     * @throws NullPointerException if the form id is null or invalid
```
kmadej: this method never throw NPE because you are catching this exception

commcare/src/main/java/org/motechproject/commcare/pull/CommcareFormImporterImpl.java
```diff
@@ -64,6 +65,20 @@ public int countForImport(Range<DateTime> dateRange, String configName) {
     }
 
     @Override
+    public boolean checkFormIdForImport(String formId, String configName) {
+        LOGGER.info("Checking form with id {} [config: {}]", formId, configName);
+        try {
+            CommcareForm form = formService.retrieveForm(formId, configName);
+            LOGGER.info("Form with id {} exists", form.getId());
+            return true;
+        } catch (NullPointerException e) {
```
kmadej: catching NPE is always bad idea, you should rather modificate retrieveForm method from formService to return null if form does not exist and then "return form != null'

kmadej (changes requested):
The feature seems to work correctly, please adjust code with my feedback

kmadej (approved):

PR #482, opened 2016-11-22
MOTECH-3031 Added Import Cases functionality in Commcare module

shanmukhm:
Cases can be imported either by given Case Id or by given date range of modification
Separate tab is added by name 'Import Cases'
User can select the basis of import i.e. 'by Id' and 'by Modification date'
User can input uuid of the case
Events are triggered for each case that is imported

### DIFF
--- a/commcare/src/main/java/org/motechproject/commcare/client/CommCareAPIHttpClient.java
+++ b/commcare/src/main/java/org/motechproject/commcare/client/CommCareAPIHttpClient.java
@@ -184,7 +184,7 @@ public class CommCareAPIHttpClient {
      * @return the JSON string representation of the cases
      */
     public String casesRequest(AccountConfig accountConfig, CaseRequest caseRequest) {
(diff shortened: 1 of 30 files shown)


kmadej:
Hi, Joanna Bebak from our QA team has tested this feature :  "I tested the feature, and when the case ID is correct, everything works correctly. Yet when one enters an incorrect case ID, the "Import complete" window appears either way, and Null Pointer Exception is visible in the logs. Once it even happened that I got not only an NPE in the logs, but also the task ran successfully twice despite an incorrect ID. As far as case import by date range is concerned, an incorrect number of the cases is imported; e.g. once, when within a given date range only two cases should have been imported, 9 have been imported instead"

shanmukhm:
Hi @kmadej, please don't test this. I still have to make changes. I'll let you know once I'm done.

shanmukhm:
Hi, you can test this.

kmadej:
Karolina left comments on JIRA

shanmukhm:
It is ready for review

PR #484, opened 2016-11-23
MOTECH-3053: Add Ability To Lookup By Other Identifiers.

pkornowski:


openmrs/src/main/java/org/motechproject/openmrs/service/OpenMRSPatientService.java
```diff
@@ -75,6 +75,17 @@
     Patient getPatientByMotechId(String configName, String motechId);
 
     /**
+     * Returns the patient with the given {@code otherName} and {@code otherId}. Configuration with the given {@code configName} will be used
```
kmadej: we should not use the word 'other', it's not obvious what it means. Just name it 'identifier' without other

openmrs/src/main/java/org/motechproject/openmrs/service/impl/OpenMRSPatientServiceImpl.java
```diff
@@ -122,6 +122,11 @@ public Patient getPatientByMotechId(String configName, String motechId) {
     }
 
     @Override
+    public Patient getPatientByOtherIdentifier(String configName, String otherId, String otherName) {
```
kmadej: getPatientIdentifier

kmadej (changes requested):

kmadej (approved):

PR #492, opened 2016-11-28
MOTECH-3050: Added transactions to dhis2 module

kmadej:


pkornowski:
@kmadej  looks good to me.

PR #498, opened 2016-12-05
MOTECH-3049: Fixed DHIS2 sync problems.

kmadej:
Send single get request if name is empty


pkornowski:
I checked this locally and it works correctly.

PR #502, opened 2016-12-06
MOTECH-3071: Changed handling DHIS2 actions from events to OSGI services

kmadej:


pkornowski:
@kmadej The code looks good to me.

PR #503, opened 2016-12-07
MOTECH-3070: Fixed parse period issues in dhis2

kmadej:


pkornowski:
@kmadej Looks good to me.

PR #506, opened 2016-12-13
MOTECH-3080: Added possibility to create observation groups using createEncounter

kmadej:


pkornowski:
@kmadej  Looks good to me.

PR #512, opened 2016-12-28
MOTECH-3049: Fixed: NPE during a 'sync'.

pkornowski:


dhis2/src/main/java/org/motechproject/dhis2/service/impl/SyncServiceImpl.java
```diff
@@ -110,6 +110,12 @@ private void addDataSets() {
         List<DataSetDto> dataSetDtos = dhisWebService.getDataSets();
 
         for (DataSetDto dataSetDto : dataSetDtos) {
+            if (dataSetDto.getDataElements() == null) {
```
kmadej: do this in createFromDetails method and adjust unit test in DataSetServiceTest

UPDATE: or add condition to getter

kmadej (changes requested):

kkedys:
@kmadej, the pull request is checked.

kmadej (approved):

PR #513, opened 2016-12-29
MOTECH-3077 CommCare Case Update - Owner Id should not be required

srujan92:
Removed the 'owner id' parameter from update case related methods.

### DIFF
--- a/commcare/src/main/java/org/motechproject/commcare/tasks/builder/CaseActionBuilder.java
+++ b/commcare/src/main/java/org/motechproject/commcare/tasks/builder/CaseActionBuilder.java
@@ -120,6 +120,14 @@ public class CaseActionBuilder implements ActionBuilder {
             parameters.add(builder.createActionParameterRequest());
 
             builder = new ActionParameterRequestBuilder()
+                    .setDisplayName(DisplayNames.OWNER_ID)
+                    .setKey(EventDataKeys.OWNER_ID)
+                    .setType(UNICODE.getValue())
+                    .setRequired(false)
+                    .setOrder(order++);
+            parameters.add(builder.createActionParameterRequest());
+
+            builder = new ActionParameterRequestBuilder()
                     .setDisplayName(DisplayNames.CLOSE_CASE)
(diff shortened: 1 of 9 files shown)


motech-gerrit:

motech-gerrit:

commcare/src/main/java/org/motechproject/commcare/tasks/builder/CaseActionBuilder.java
```diff
@@ -120,14 +120,6 @@ public CaseActionBuilder(CommcareConfigService configService) {
             parameters.add(builder.createActionParameterRequest());
 
             builder = new ActionParameterRequestBuilder()
-                    .setDisplayName(DisplayNames.OWNER_ID)
-                    .setKey(EventDataKeys.OWNER_ID)
-                    .setType(UNICODE.getValue())
-                    .setRequired(true)
```
kmadej, line 126: why did you remove ownerId ? It should be still available, however not required. So try to set required to false

kmadej (changes requested):

commcare/src/main/java/org/motechproject/commcare/tasks/builder/CaseActionBuilder.java
srujan92: Since we are not using owner id in update case, I thought parameter is not needed. Ok, I will revert the changes and set required to false.

srujan92 (commented):
@kmadej please see the update changes and see if this is what the required solution for this issue.

kmadej:
@srujan92 it seems ok, please fix tests

kmadej:
@srujan92 please change line 531 in CommcareTasksIntegrationBundleIT. Now it should be '.setRequired(false)'

kmadej (approved):

srujan92:
@kmadej looks like the building gods have granted us successful build :-) , please merge the code when its feasible to you.

kmadej:
@srujan92 well said :) I've merged the code

PR #515, opened 2017-01-03
MOTECH-3049: Fixed problem with transactions.

pkornowski:


kkedys:
@kmadej When the DHIS2 module is synchronized it is possible to change the tab

PR #520, opened 2017-01-25
MOTECH-3118: Fixed: Cannot Save Additional Identifier to Patient.

pkornowski:


kkedys:
@kmadej, the issue is fixed.

PR #523, opened 2017-01-27
MOTECH-3118: Deleted unnecessary Transactional annotations.

pkornowski:


kkedys:
The problem still exists.

PR #527, opened 2017-02-09
Remove motech.version and motech.groupId properties

kmadej:
These properties are redundant, because we can use project.version and project.groupId.
Additionally, using them are problematic because during release, travis/jenkins
does not bump up it automatically.


tools/version_update.py
```diff
@@ -5,7 +5,7 @@
 pom = open("pom.xml")
 dom = parse(pom)
 pom.close()
-mv = dom.getElementsByTagName('motech.version')
+mv = dom.getElementsByTagName('project.version')
```
mkruszynski: I think that this script is now unnecessary, it was only creating a commit with bump up of _motech.version_ property value.

mkruszynski (changes requested):

mkruszynski:
@kmadej looks good to me

mkruszynski (approved):

PR #528, opened 2017-02-10
MOTECH-3114: Added bubble parent ID to commcare case datasource.

pkornowski:


kkedys:
@kmadej, "Parent ID" bubble works correctly.

PR #529, opened 2017-02-14
MOTECH-3133: Added a new way to create Commcare case properties

mkruszynski:


kkedys:
@kmadej the issue is fixed.

PR #532, opened 2017-03-08
MOTECH-3144: Added new Observation Lookups and Unit tests.

pkornowski:
Added lookups:

- By Patient ID, Concept ID and Value
- By Encounter ID and Concept ID
- By Encounter ID, Concept ID and Value

Removed lookup:

- By Patient ID and Value


kkedys:
@pkornowski the new lookups are added and they works correctly.

PR #533, opened 2017-03-09
MOTECH-3150: Fixed npe during create commcare case.

pkornowski:


kkedys:
@pkornowski The ticket is fixed.

PR #535, opened 2017-03-27
MOTECH-3156: Added a possibility to use atom feeds from OpenMRS server without Atom Client module.

pkornowski:


kkedys:
@pkornowski, the option work correctly, but on OpenMRS module, the scrollbar missing by the way, one can not see all the options in the configuration (especially those added in this ticket).

kkedys:
@pkornowski Thanks for add the scrollbar. Now all work correctly.

PR #536, opened 2017-03-27
MOTECH-3161: Added update observation action to the OpenMRS.

pkornowski:


kkedys:
@pkornowski Update observation action works correctly.

PR #537, opened 2017-03-27
MOTECH-3160: Added a possibility to get observation by concept as values.

pkornowski:


kkedys:
@pkornowski the issue is resolved.

PR #539, opened 2017-03-30
MOTECH-3156: Fixed scheduler issue and added unit test.

pkornowski:


kkedys:
Checked. The issue is fixed.

PR #540, opened 2017-04-05
MOTECH-3169: Added post action parameter, uuid for Create/Update observation action.

pkornowski:


kkedys:
Checked. It works correctly.

PR #541, opened 2017-04-05
MOTECH-3168: Added incrementing current page id after every page reading.

pkornowski:


kkedys:
Checked. Everything works correctly.

PR #542, opened 2017-04-10
MOTECH-3170: Added deserializer to OpenMRS Observation.

pkornowski:


kkedys:
Checked. everything works correctly.

PR #546, opened 2017-04-21
MOTECH-3178: Added logging request and response with http code to ope…

pkornowski:
…nmrs actions.


kkedys:
When the task fails after request is sent, there should be visible status code and response message.

kkedys:
Checked. Now it works correctly.

PR #549, opened 2018-04-05
Fixed Batch ITs

kmadej:
In JobServiceBundleIT new job was created with past date and the test
failed. I changed the date, so currently the job is properly created via
motech scheduler.


kmadej:
Changes have been implemented in another PR - https://github.com/motech/modules/pull/548

